Issue #96, opened 2013-06-19
Add Deploy/Revert Hooks


theory:
Add the ability to run arbitrary scripts after a `sqitch deploy` and after a `sqitch revert`. These might be used to, for example, update schema permissions or ownerships in a bulk way. Or maybe they would `VACUUM` and `ANALYZE` a Postgres database.

There should also be a way to disable them via a command-line option (and configuration?) to `deploy` and `revert`. They also should not run when `--log-only` is specified, I think.

Suggested by @bdigital.

ngmaibulat:
Chat GPT mentioned hooks: https://chat.openai.com/share/85486807-0e15-4130-a5b6-dd2626aa0632

placetobejohan:
> Chat GPT mentioned hooks: https://chat.openai.com/share/85486807-0e15-4130-a5b6-dd2626aa0632

I ran into the same rabbit hole 🤦 Would be a nice feature though.

theory:
Never trust a bot.

fluca1978:
> Thinking back, I believe what I envisioned was SQL scripts, not arbitrary commands. It likely arose as I developed verify scripts and [templating additional scripts](https://justatheory.com/2014/01/templating-tests-with-sqitch/) (though it looks like that blog post was 6 months after I created this issue). In all honestly I don't know what that'd gain us vs. just adding more lines to a deploy or revert script.

I was not aware of this way of templating, I think this surely adds modularity.

> 
> So perhaps it makes as much or more sense to add the ability to run arbitrary scripts, as you sketch out, [@fluca1978](https://github.com/fluca1978). So much the better if they don't have to come after a specific change, just at the end (or start?) of a deploy or revert.
> 

Ok, I can work *with no rush* on it in the next weeks and provide a PoC or some kind of implementation of this.


> In that vein, I tend to think that the configuration would be in the `[deploy]` and `[revert]` configs (and related configs like rework). Something like
> 
> [deploy.hooks]
>     before = /bin/sh -c 'echo hello'
>     after = /bin/sh -c 'echo goodbye'
> 
> And then of course one could turn them on or off on a per-engine an per-target basis.


Uhm, I think the configuration should be simple because the more we add capabilities, the more it becomes complex to configure and to "parse". Is the following a good starting point?

```
[deploy.hooks]
   enabled = on   # enabled hooks if not specified otherwise
  before = psql -c 'select current_date;'
  after = dbicdump

[revert.hooks]
   enabled = off
[...]

PR #178, opened 2014-08-28
pull request for plugin support #96

cazador481:
I have created a plugin support for the engine.  This goes to issue #96.
I also created a plugin that will run an executable in the following directories:
after_verif, after_deploy,after_revert


fluca1978:
I'm slowly working on a PoC about this, and here are some thoughts.
In the case the work is worth, I can continue and start showing some code.

## Architecture

Do not use plugins, simply *inject* the hooks into the `run_deploy` and friends.
In particular, provide a single method (private?) that can be called for instance as:
```
	    $self->run_hook( action => q/deploy/,
			     at     => q/after/,
			     change => $change );
```

so that the method knows if it is called for a deploy/revert (`action`), when (`at`) meaning, after or before, and about which change. This clearly resembles an SQL trigger definition.

## Configuration

I see a configuration like the following:

```
% cat sqitch.conf
[target "test"]
        uri = db:pg://biblio:xyz@rachel/bibliodb_test
        hooks.enabled = on
        hooks.after.deploy.command = echo within target

[target "prod"]
        uri = db:pg://biblio:xyz@rachel/bibliodb

[core]
         engine = pg
        hooks.enabled = off
        hooks.after.deploy.command = echo within core

```

There are:
- `hooks.enabled` that defines if hooks are supported
- `hoosk.<when>.<action>.command` that define which command to execute

The target variables override the core ones, so you can quickly configure a single set of variable for all targets but disable for a specific one or use a different command for a particular target in the set.
In the previous example, only `test` has an hook enabled, while `prod` has not.


## Problems and doubts
- we should provide at least two commands, or a way to indicate what type of command to execute. A command that uses a database connection (e.g., to perform schema ownership changes) and one to execute an external command. Moreover, this raises a problem of order in execution.
- executing external scripts (i.e., outside the database) poses security risks, so we should at least check the user provides an absolute path to a command;
- according to me, executing hook scripts should never rollback a deployed change, this is an *extra* action performed outside the sqitch control flow;
- do we need to define an execution chaining prevention? What if I re-deploy multiple changes, do we need to execute the scripts one for every deploy or one at the end (this resemble an SQL trigger for row or for statement);
- every external command will automatically receive a set of parameters, like the target name, the change id, the action, the firing time, etc., so that it can perform some introspection by its own;
- I believe we should introduce a specific `Hook` class in the Sqitch hierarchy to better handle the configuration, in my PoC I've not;
- as an *extra*: a hook command should not be able to fire up a sqitch process...

@theory and @cazador481 thoughts?

PR #182, opened 2014-09-22
Update mysql.sql: DATETIME(6) is not compatible with MySQL < 5.6.4

robertblank:
http://dev.mysql.com/doc/refman/5.6/en/date-and-time-type-overview.html

"MySQL 5.6.4 and up permits fractional seconds for TIME, DATETIME, and TIMESTAMP values, with up to microseconds (6 digits) precision. To define a column that includes a fractional seconds part, use the syntax type_name(fsp), where type_name is TIME, DATETIME, or TIMESTAMP, and fsp is the fractional seconds precision. For example:

CREATE TABLE t1 (t TIME(3), dt DATETIME(6));"

This means that for versions MySQL < 5.6.4 the syntax dt DATETIME(6) produces an error.
As this is an optional value I suggest to use DATETIME without fsp to keep it more compatible.

### DIFF
--- a/lib/App/Sqitch/Engine/mysql.sql
+++ b/lib/App/Sqitch/Engine/mysql.sql
@@ -7,7 +7,7 @@ CREATE TABLE projects (
                     COMMENT 'Unique Name of a project.',
     uri             VARCHAR(255) NULL UNIQUE
                     COMMENT 'Optional project URI',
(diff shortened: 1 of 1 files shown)


theory:
FYI, I went ahead and brought the minimum down to 5.5.0 in e2beb09. See #218 for details.

PR #223, opened 2015-04-10
Update sqitchtutorial-oracle.pod

hartzell:
A couple of small typo fixes.

I'm unsure about the deletion around line 518.  You say above that `squitch` automatically inserts the error handling code, and you have *not* included it in the example, so I think that the prose suggesting that you did (and one should) is a left-over.


theory:
Yeah, I think that line was left over from an earlier iteration. Thanks!

PR #225, opened 2015-04-10
Update sqitchtutorial-oracle.pod

hartzell:
Either the prose or I are confused about how a SID is used in the example about deploying to production.


theory:
Great feedback, thanks so much for the careful read and PRs!

PR #228, opened 2015-04-20
Add a documentation note about the MySQL::Config requirement to read my.cnf files

zostay:
I had trouble figuring out why my system wasn't reading my.cnf files as documented. Reading the code, I realized I was missing an optional dependency. Added a doc to help others with this problem in the future.


theory:
Thank you!

PR #247, opened 2015-08-11
Registry prefix feature

stefansbv:
The Oracle engine needs a review and testing.


stefansbv:
There are some problems with the MySQL engine, yesterday I thought that the error is not important, but now I see that the feature doesn't work. Sorry...

stefansbv:
The registry-prefix feature for the Firebird and the MySQL engines are now working, but the implementation details can be improved, probably.

rzane:
:+1: Looking forward to this!

hiqo-amaksimovich:
Hello,
Is there any chance to get the prefix change in?
Regards,

theory:
Yes, eventually, but I personally don't have time to work on Sqitch right now. @stefansbv, maybe you'd like to rebase it, though?

stefansbv:
Yes, I can try to rebase it.

stefansbv:
Not finished, work in progress...

stefansbv:
There is a problem with the `firebird` test. Sometimes it freezes after the `ok 79 - live database` line (the CPU usage goes to 100%). I don't think is perl version related but it is random. I also noted this problem in the master branch, so it must be an older issue.

theory:
Yeah, I just re-run the test that fails with that error. I don't know where the issue lies, either.

theory:
@stefansbv Got time to resurrect this PR? Rebase it on master and try to get it working again? Thank you!

stefansbv:
Weird, I was expecting lots of failures from oracle and firebird but not from mysql... :-)

theory:
Yeah, I just manually re-run the Travis build that fails with Firebird freezes. It usually passes on the next run. I chalk it up to Firebird and don't sweat it.

lib/App/Sqitch/Engine/firebird.pm
```diff
@@ -621,7 +621,7 @@ sub name_for_change_id {
               JOIN $tags t ON c2.change_id = t.change_id
              WHERE c2.committed_at >= c.committed_at
                AND c2.project = c.project
-        ), '@HEAD')
+        ), '\@HEAD')
```
theory: @ does not interpolate in single-quoted strings.

theory:
It [looks like](https://travis-ci.org/theory/sqitch/jobs/297175268#L2276) an instance of `:prefix` isn't getting replaced in the MySQL file.

lib/App/Sqitch/Engine/firebird.pm
stefansbv: Except when is inside a `qq{}` ?!

theory: Oh, heh, yes.

stefansbv:
I tried to debug the `firebird` engine test, I think is an issue with `DBD::Firebird`, not the `Firebird` server:

[Sqitch firebird test failure ](https://gist.github.com/stefansbv/9f57e627007aea56f6ecbe08daa190e5)

theory:
Oh interesting. Can you reduce it? Or anyway point @mariuz at it?

theory:
<nag/>

mariuz:
@theory we skip for the moment memory leaks long test in DBD::Firebird https://github.com/mariuz/perl-dbd-firebird/blob/master/t/60-leaks.t
To run them you can add  MEMORY_TEST=1 DBI_PASS=masterkey make test

theory:
@stefansbv When that test hangs, I just rerun the test and it passes. You can ignore it, I think; it's in all the branches, no?

@mariuz I'm sorry, I don't understand what you're suggesting. Do you want us to try to skip some tests that cause a hang? I'm not sure that would be easy to do…

stefansbv:
@theory, yes, I think it can be ignored, I did not encounter problems when I was using the app with this feature enabled.

The worst alternative is to remove the support for Firebird from Sqitch.

theory:
Meh, it's been like that for years. I figure it goes away once @mariuz figures out how to fix the issue.

PR #260, opened 2015-09-16
Oracle Wallet Support

tprocter:
In shifting to use Oracle wallets for improved security, we found that sqitch almost but not quite supports connectivity strings needed for wallets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 .build/
 *.mo
 .al
+*.komodoproject
--- a/lib/App/Sqitch/Engine/oracle.pm
+++ b/lib/App/Sqitch/Engine/oracle.pm
@@ -667,16 +667,22 @@ sub _script {
         $conn .= qq{/"$pass"};
     }
     if (my $db = $uri->dbname) {
-        $conn .= '@';
-        $db =~ s/"/""/g;
-        if ($uri->host || $uri->_port) {
-            $conn .= '//' . ($uri->host || '');
-            if (my $port = $uri->_port) {
-                $conn .= ":$port";
+        if ( ( ! $conn ) and ( ! $uri->host ) ) {
+            # OS authentication or Oracle wallet (no username or password)
+            $conn = '/@' . $uri->dbname;
+		}
+		else {
+            $conn .= '@';
+            $db =~ s/"/""/g;
+            if ($uri->host || $uri->_port) {
+                $conn .= '//' . ($uri->host || '');
+                if (my $port = $uri->_port) {
+                    $conn .= ":$port";
+                }
+                $conn .= qq{/"$db"};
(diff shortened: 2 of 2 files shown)


theory:
Thank you! Do you feel like adding something about using Oracle Wallets to [sqitch-passwords](https://github.com/theory/sqitch/blob/master/lib/sqitch-passwords.pod)? Maybe in the "Don't use Passwords" section?

tprocter:
will do.  ETA tomorrow.

theory:
Please confirm that Oracle wallet still works after c0767e8.

theory:
Make that after 3d61dcc.

tprocter:
authority->() is returning ":@" for this config, so it's not hitting the right block (probably a parsing issue of the URI).

example URI: db:oracle://:@/wallet_tns_name

The sqlplus command needs to be "/@wallet_tns_name"

theory:
Does `db:oracle:///wallet_tns_name` work?

tellier:
Hi 

sqitch config --system target.tns.uri 'db:oracle:///secure_user_tns.tpg'  doesn't work I get the following error 
ORA-12154: TNS:could not resolve the connect identifier specified (DBD ERROR: OCIServerAttach)

tprocter:
So, the above format doesn't work in DBD since the @ is part of the resource name.

theory:
Should it be `db:oracle:///@secure_user_tns.tpg`, then?

tellier:
that syntax gives me the following error 
ORA-12154: TNS:could not resolve the connect identifier specified (DBD ERROR: OCIServerAttach)

theory:
Okay, let's try this: Please tell me exactly the syntax required both for DBD::Pg and for SQLPlus* and I'll add some tests and get it to work.

tellier:
sqlplus /@secure_user_tns.tpg

DBD::Oracle is as follows:
my $dbh = DBI->connect("dbi:Oracle:secure_tns_entry.here", '', '');

theory:
Uh, those are different strings. Should the DBD::Oracle one be dbi:Oracle: secure_user_tns.tpg` to match sqlplus?

tellier:
Yes, this should be clearer:
sqlplus:
connect /@secure_user_tns.tpg

DBD:
$dbh = DBI->connect("dbi:Oracle:secure_user_tns.tpg", '', '');

theory:
@tellier Okay, I think I got it in ba07968. Would you mind trying again?

tellier:
Hi David, 

Sorry for the delay looks like this change works correctly I was able to access the DB with the wallets and revert, deploy, verify and status where run..

theory:
Awesome, thank you! Will do a release soon.

PR #265, opened 2015-10-11
Use $Config{siteprefix} as default directory for etc.

### LINKED ISSUE #264: Install PREFIX for etc directory is not the same as the rest of the distribution
When I try to install Sqitch with:

    cpanm App::Sqitch

It install most files under the `/usr/local` PREFIX as expected. But the `etc` directory gets incorrectly installed under `/usr/etc` instead of `/usr/local/etc`. Any idea why the `/usr/local` PREFIX is not honored in this case?. How does the package figures out where should the `etc` directory go?.

In my system (Debian 8) there is no need to be root in order to install Perl packages, the staff group has all the permissions required to do it. But in this case, the installation fails due to permission problems because Sqitch is trying to write to `/usr/etc` instead of `/usr/local/etc`.

Thanks for your time.

carragom:
Fixes #264. This will use `$Config{siteprefix}` instead of `$Config{prefix}` to place the `etc` directory on a default install, to be consistent with the rest of the files. Also it will give priority to the value of `install_base` over `prefix` since it seem to be the recommended way in [Module::Build's](https://metacpan.org/pod/Module::Build#prefix) docs.

### DIFF
--- a/inc/Module/Build/Sqitch.pm
+++ b/inc/Module/Build/Sqitch.pm
@@ -35,6 +35,16 @@ sub new {
 
 sub _getetc {
     my $self = shift;
+    my $prefix;
+
+    if ($self->installdirs eq 'site') {
+        $prefix = $Config::Config{siteprefix} // $Config::Config{prefix};
+    } elsif ($self->installdirs eq 'vendor') {
+        $prefix = $Config::Config{vendorprefix} // $Config::Config{siteprefix} // $Config::Config{prefix};
+    } else {
+        $prefix = $Config::Config{prefix};
+    }
+
     # Prefer the user-specified directory.
     if (my $etc = $self->etcdir) {
         return $etc;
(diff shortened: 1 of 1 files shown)


theory:
I think it might make sense to check the Module::Build `installdirs`attribute and select the Config key appropriately. Something like:

    my $which = $self->installdirs;
    $which = '' if $which eq 'core';
    my $prefix = $self->installdirs eq 'core'   ? $Config{prefix}
               : $self->installdirs eq 'site'   ? $Config{siteprefix}   || $Config{prefix}
               : $self->installdirs eq 'vendor' ? $Config{vendorprefix} || $Config{siteprefix} || $Config{prefix};

carragom:
Makes sense. Here is another stab at it. Code is a little different but should achieve the same. Let me know what you think. By the way, thanks for the great piece of software.

PR #266, opened 2015-10-16
documentation fix

dredozubov:
There were two `REVERT`s instead of `COMMIT` in the docs. Also vim stripped whitespaces at the end of the lines and left it like this, it seems much nicer this way.

[Review on Reviewable](https://reviewable.io/reviews/theory/sqitch/266)

### DIFF
--- a/lib/sqitchtutorial-mysql.pod
+++ b/lib/sqitchtutorial-mysql.pod
@@ -34,7 +34,7 @@ Usually the first thing to do when starting a new project is to create a
 source code repository. So let's do that with Git:
 
   > mkdir flipr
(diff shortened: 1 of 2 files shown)


dredozubov:
Not sure what to make out of perl 5.20 build fail. I'm not a perl guy. It seems unrelated to the PR.

theory:
Thank you! I pulled in this change in 3eec3ab. Note that a bunch of the trailing space you removed was there by design. Some wan't, though, so I removed those. 

Not sure what's up with the JSON issue on the 5.20 build. I'll keep an eye on it.

PR #292, opened 2016-04-27
Fix spelling for PostgreSQL

slapresta:
This fixes a few instances on the page where "PostgreSQL" was mistakenly spelled as "PostreSQL".

Have a good day!


theory:
Well that was embarrassing. Thank you!

PR #300, opened 2016-06-01
Add Redshift Engine

stevecaldwell77:
Adds a new engine for AWS Redshift.  This code should be considered alpha code - I've tested deploy/verify/revert against my database, but there could very well be edge cases that are not working.

I'm also very open to suggestions on ways to improve the code.

NOTE: this depends on URI::db to support redshift, see: https://github.com/theory/uri-db/pull/12

### DIFF
--- a/lib/App/Sqitch/Engine/pg.pm
+++ b/lib/App/Sqitch/Engine/pg.pm
@@ -268,11 +268,13 @@ sub log_new_tags {
         $sqitch->user_email
     );
 
+    my $tagcol = $self->_tag_column();
+
     $self->dbh->do(
-        q{
+        qq{
             INSERT INTO tags (
                    tag_id
-                 , tag
+                 , $tagcol
                  , project
                  , change_id
                  , note
@@ -314,8 +316,9 @@ sub log_revert_change {
     my $dbh = $self->dbh;
 
     # Delete tags.
+    my $tagcol = $self->_tag_column();
     my $del_tags = $dbh->selectcol_arrayref(
-        'DELETE FROM tags WHERE change_id = ? RETURNING tag',
+        "DELETE FROM tags WHERE change_id = ? RETURNING $tagcol",
         undef, $change->id
     ) || [];
 
@@ -374,6 +377,7 @@ sub _update_ids {
     my $plan = $self->plan;
     my $proj = $plan->project;
     my $maxi = 0;
+    my $tagcol = $self->_tag_column();
 
     $self->SUPER::_update_ids;
     my $dbh = $self->dbh;
@@ -392,15 +396,15 @@ sub _update_ids {
               FROM changes
              WHERE project = ?
         });
-        my $atag_sth = $dbh->prepare(q{
-            SELECT tag
+        my $atag_sth = $dbh->prepare(qq{
+            SELECT $tagcol
               FROM tags
              WHERE project = ?
                AND committed_at < ?
              LIMIT 1
         });
-        my $btag_sth = $dbh->prepare(q{
-            SELECT tag
+        my $btag_sth = $dbh->prepare(qq{
+            SELECT $tagcol
               FROM tags
              WHERE project = ?
                AND committed_at >= ?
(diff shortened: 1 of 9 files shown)


lib/App/Sqitch/Engine/pg.pm
```diff
@@ -106,19 +106,7 @@ has dbh => (
                 goto &hurl;
             },
             Callbacks         => {
-                connected => sub {
```
theory: Did you mean to make these changes to the pg driver?

theory: Oh, you did, sorry, please ignore that question.

theory:
I'm wondering if it's necessary to have a separate subclass for redshift. Are they really all that different? See for example how I just tweaked things to [add Postgres-XC support](https://github.com/theory/sqitch/compare/2a63bb1...c380575). Fine if the answer is yes, but I'm not really able to get a sense yet.

Also, is there a way to run automated testing in Travis CI on a redshift instance?

stevecaldwell77:
@theory:

> I'm wondering if it's necessary to have a separate subclass for redshift.

It's a good question.  I've found 8 places where Redshift needed to handle things differently than Postgres:

- registry file
- function used to get current time
- date to-string formatting
- no support for 'client_min_messages'
- "tag" is a reserved word, and therefore needs to be quoted when used as a column name
- use the listagg() function
- no support for the correlated subquery in App::Sqitch::Role::DBIEngine::changes_requiring_change()
- no support for the "RETURNING" expression.

So that would be 8 if/then's we would need to litter throughout the pg engine class.  I usually shy away from lots of conditionals in my code, but it's your library and I'm happy to conform to your preference :)

> Also, is there a way to run automated testing in Travis CI on a redshift instance?

A bit of a tough one.  You could have a script that boots up a new Redshift cluster using CloudFormation, runs tests against it, then shuts it down.  That would cost $0.25 every time you do it, so could get pretty expensive if you ran it too often.

clearclaw:
Any prognosis on this PR?  (I'd love to use Sqitch against Redshift)

theory:
Is there a way to get Travis to run tests against a redshift instance? That would make me a whole lot more comfortable just merging it in.

stevecaldwell77:
@theory:

> Is there a way to get Travis to run tests against a redshift instance?

As I noted above, that would be difficult and/or expensive.  Running an always-on redshift cluster costs at least $1,380 a year, which I'm guessing is prohibitive.

You could launch a new cluster every time Travis runs, and then destroy it when Travis completes.  This would cost $0.25 for every Travis run, and adds a long startup time to a Travis run (+5 mins to wait for a Redshift cluster to become available, in my experience).

Perhaps an acceptable solution would be to have a test that boots up and runs against a new cluster, but isn't part of a CI process (and has to be run manually).

theory:
Bah, sorry to make you repeat history there, @stevecaldwell77. My head has been elsewhere altogether. Lemme see if I can find someone to fund this.

FsecureSamiTikka:
I really would like to use sqitch to work with Redshift but I can't get this to work. This PR did not build until I rebased it to master. And even then I could not get it to work. I get what seems like an error from Redshift and it does not like the word "tag"?? Am I using it wrong? I tried to `sqitch deploy db:redshift:<name of database`. Should I install DBD::Pg (I did) or something else? (I thought Sqitch uses the psql client because it wants to know where it is.)

green2k:
Hello, 

it would be pretty cool to support Redshift engine! Is there any problem with resolving this "feature request"? Guys in previous comments probably have some knowledge about it & know the differences between PostgreSQL & Redshift, so it shouldn't be hard to implement. 

I've never used the Sqitch tool, so I even don't know the project's code.. but if needed, I can help! 

Does anybody have some forked project with Redshift support? Or have some workaround? 

Thanks!

Pali

theory:
What we really need is an easy way to test it against a RedShift database. Ideally this would be through Travis. Suggestions?

theory:
Maybe something can be borrowed from the [sqlalchemy redshift live testing](https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/commit/c3ec3ccd77b35cc0b7498b1b826345bf5b2828b5#diff-354f30a63fb0907d4ad57269548329e3) stuff?

eric-aldinger:
Have you considered LocalStack for CI testing? It mocks RedShift on a local port 
https://github.com/atlassian/localstack

theory:
Anyone got the TUITs to finish this up?

PR #301, opened 2016-06-02
Change WHERE clause to return false

dtseiler:
I'm assuming this is what you meant here.

### DIFF
--- a/lib/sqitchtutorial-oracle.pod
+++ b/lib/sqitchtutorial-oracle.pod
@@ -509,7 +509,7 @@ from the table with a false C<WHERE> clause. Add this to F<verify/users.sql>:
 
   SELECT nickname, password, timestamp
     FROM flipr.users
-   WHERE 1 = 1;
+   WHERE 0 = 1;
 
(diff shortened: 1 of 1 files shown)


theory:
Nope. By making it `WHERE 1 = 1`, it should never scan any rows. If you make it `WHERE 0 = 1`, it will do a table scan. Since the whole point is to make sure the table exists with the expected columns, a table scan is unnecessary, so I avoid it.

dtseiler:
Not quite sure I follow you. If you have WHERE 1=1, you're saying you want all the rows. Why wouldn't it then scan the table?

theory:
Oh. Yeah, I'm an idiot, sorry.

dtseiler:
No worries ;)

dtseiler:
PS - I don't know if you have similar examples in the other tutorials that might need similar corrections. I'm an Oracle DBA so I was only reading that one.

theory:
No, the others are able to use clearer expressions, like `WHERE FALSE`.

PR #302, opened 2016-06-14
update registry configuration demo for mysql tutorial

pengfei-xue:
seems like the latest version moved registry configuration to be under target section, caused lots of time to track it :(, this doc should be updated.


pengfei-xue:
by the way, my sqitch version is `0.996`

theory:
It does not create the target like that. I just ran it, and, other than the `client` path being incorrect (fixed in e07c46a), the configuration is exactly as you see in the tutorial. I think your version of Sqitch might be very old: the latest is v0.9994; 0.996 is 18 months old.

pengfei-xue:
@theory ok, thanks for the explanation :)

PR #304, opened 2016-06-18
Use DateTime->set_locale instead of ->set to change the locale

autarch:
Requires DateTime 1.04+

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,9 @@
 Revision history for Perl extension App::Sqitch
 
+     - Use DateTime->set_locale instead of ->set to set the locale. Using
+       ->set may actually change the local time unintentionally. Patch by Dave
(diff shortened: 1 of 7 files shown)


autarch:
Note that using `->set` this way will start warning in a future version of DateTime

theory:
Thanks Dave!

Changes
```diff
@@ -1,5 +1,9 @@
 Revision history for Perl extension App::Sqitch
 
+     - Use DateTime->set_locale instead of ->set to set the locale. Using
+       ->set may actually change the local time unintentionally. Patch by Dave
+       Rolsky. GitHub PR #303.
```
ilmari: ITYM #304.

theory: Yep, thanks, fixed in 5015a3f.

PR #307, opened 2016-07-05
Allow non-superuser to create sqitch.

djk447:
I'm rather careful to not use superusers when I don't have to. I have been able to avoid needing superuser status for deployments by giving users ownership over their own schemas etc. I ran into trouble deploying sqitch because it would error if the schema was already created.

The basic use case is, as a superuser, I would like to be able to run: 
CREATE SCHEMA sqitch AUTHORIZATION sqitch_user;

then when connected as sqitch_user, I want sqitch to deploy as normal, the problem is it errored completely if the schema already exists. If it exists, I would prefer it continue. If I had failed to give it ownership of the schema, further bits would fail (commenting, creating tables etc.) but this allows for this use case to work and for the sqitch_user not to need superuser privileges. (I can grant it privileges on other schemas that it needs to work in otherwise).

Let me know if there are any troubles with this.

### DIFF
--- a/lib/App/Sqitch/Engine/pg.pm
+++ b/lib/App/Sqitch/Engine/pg.pm
@@ -157,7 +157,7 @@ sub initialized {
     my $self = shift;
     return $self->dbh->selectcol_arrayref(q{
         SELECT EXISTS(
-            SELECT TRUE FROM pg_catalog.pg_namespace WHERE nspname = ?
+            SELECT TRUE FROM pg_catalog.pg_tables WHERE schemaname = ? AND tablename = changes
         )
     }, undef, $self->registry)->[0];
 }
--- a/lib/App/Sqitch/Engine/pg.pm
+++ b/lib/App/Sqitch/Engine/pg.pm
@@ -157,9 +157,9 @@ sub initialized {
     my $self = shift;
     return $self->dbh->selectcol_arrayref(q{
         SELECT EXISTS(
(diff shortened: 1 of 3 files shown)


djk447:
Well, this doesn't work as the only thing that

theory:
Well, we could change it to check to see if the changes table exists in that schema. [See `initialized()`](/theory/sqitch/blob/master/lib/App/Sqitch/Engine/pg.pm#L156).

djk447:
Okay. I changed that and I think that should work now. Thanks!

lib/App/Sqitch/Engine/pg.pm
```diff
@@ -157,7 +157,7 @@ sub initialized {
     my $self = shift;
     return $self->dbh->selectcol_arrayref(q{
         SELECT EXISTS(
-            SELECT TRUE FROM pg_catalog.pg_namespace WHERE nspname = ?
+            SELECT TRUE FROM pg_catalog.pg_tables WHERE schemaname = ? AND tablename = changes
```
theory: `changes` needs single quotes around it, `'change'`. Or better make it a `?` and pass the string "changes":

    return $self->dbh->selectcol_arrayref(q{
        SELECT EXISTS(
            SELECT TRUE FROM pg_catalog.pg_namespace WHERE nspname = ? AND tablename = ?
        )
    }, undef, $self->registry, 'changes')->[0];

djk447:
Thanks. Done.

theory:
Finally worked this in. See also  5bd23c59978. Thank you!

Issue #328, opened 2017-04-18
Should we add the core.pager option to override the PAGER environment variable?


yati-sagade:
I noticed that running `sqitch log` produces ANSI color codes, which is sent to `less` on my box by `IO::Pager`. This does not have the intended effect of course, and doing `export PAGER=less -r` resolves it. However, might people not want to set a pager for sqitch alone? If we think so, here's a proposal:

```bash
$ sqitch config --user core.pager 'less -r'
```

That already works! Then, this small patch (or anything that has its effect) will do the trick:

```diff
diff --git a/lib/App/Sqitch.pm b/lib/App/Sqitch.pm
index 77924df..462c02b 100644
--- a/lib/App/Sqitch.pm
+++ b/lib/App/Sqitch.pm
@@ -187,6 +187,9 @@ sub go {
             args    => $cmd_args,
         });
 
+        local $ENV{PAGER} = $config->get(key => "core.pager") //
+            $ENV{PAGER};
+
         # 7. Execute command.
         $command->execute( @{$cmd_args} ) ? 0 : 2;
     } catch {
```

Let me know if you think this is worth implementing (IMO it is a quick win, but you may not want a config creep, but again, sqitch UI is inspired by git's, so...)

PS: I'm playing with sqitch as part of this month's CPAN PRC -- congratulations for the really well written code :)

theory:
Thanks! I like it! Care to write a test, document it in `lib/sqitch-config.pod`, and submit a PR?

theory:
Done in #329.

PR #329, opened 2017-04-21
Implement the core.pager config for overriding the system PAGER.

yati-sagade:
As discussed in #328, this PR adds support for the `core.pager` setting prop. I thought about what to do when the config file ends up having multiple values for the pager property, and IMO dying would be a bit extreme (imagine `sqitch log` failiing with `Multiple values for key "core.pager"`). Happy to change it if we think some other strategy would be appropriate.

This is my CPAN PRC April 2017 assignment.


theory:
Hrm. This is good, but looking for precedents, I think the handling of the `EDITOR` environment variable is the one to follow. To quote changes from the last release:

>     - Fixed editor selection to prioritize the `core.editor` configuration
>       variable over the `$EDITOR` environment variable. The `$SQITCH_EDITOR`
>       environment variable still trumps all. Thanks to Jim Nasby for the pull
>       request (#296).

You can see the precedence defined [here](/theory/sqitch/blob/master/lib/App/Sqitch.pm#L127). Can you update your PR to follow that same pattern, adding a `SQITCH_PAGER` environment variable?

yati-sagade:
Done! :)

theory:
This looks great. However, in no other place has the code bothered to worry about whether a config key has multiple values. For example, the editor attribute just looks like this:

    default => sub {
        return
             $ENV{SQITCH_EDITOR}
          || shift->config->get( key => 'core.editor' )
          || $ENV{VISUAL}
          || $ENV{EDITOR}
          || ( $^O eq 'MSWin32' ? 'notepad.exe' : 'vi' );
    }

So do we really need the `get_last_value` method? Or should that perhaps be considered a better way to always fetch a single value? Or maybe there should be a config method to complain if a key has more than one value?

I realize that's all orthogonal to the raison d'être for this patch, and I'm happy to merge it either way. If we do decide to keep `get_last_value`, you should document it in order to get the [tests](https://travis-ci.org/theory/sqitch/jobs/224687198) passing again.

yati-sagade:
Hey, I think you're right about the orthogonality of these two issues, and I've updated the PR to just use `$config->get`, like for `core.editor`. I think a separate PR for warning when multiple values are present, while addressing this for all config keys makes more sense. It can be as simple as rigging `App'Sqitch'Config` to catch the error from `Config'Gitlike` and then die with a more descriptive error message, pointing the user to the faulty config file + line.

theory:
Thanks!

PR #332, opened 2017-05-24
remove `no Moo::sification;`

perigrin:
Enforcing the Moo-only-ness of `App::sqitch` leads to killing use cases where using `App::Sqitch` programmatically from inside (for example) a test suite to set up the test environment failed if at any point something (say Catalyst) was to load Moose previously.

### DIFF
--- a/dist/sqitch.spec
+++ b/dist/sqitch.spec
@@ -40,7 +40,6 @@ BuildRequires:  perl(Locale::TextDomain) >= 1.20
 BuildRequires:  perl(Module::Build) >= 0.35
 BuildRequires:  perl(Moo) >= 1.002000
 BuildRequires:  perl(Moo::Role)
-BuildRequires:  perl(Moo::sification)
 BuildRequires:  perl(namespace::autoclean) >= 0.16
 BuildRequires:  perl(parent)
 BuildRequires:  perl(overload)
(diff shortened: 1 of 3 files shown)


theory:
I had put that in there to ensure that I really go rid of all traces of Moose. Perhaps you could add a test to do that? I just want to be sure Moose doesn't creep back in.

Thanks!

perigrin:
Yep, understood. But it completely stops a valid use case. This should add a test equivalent to what was there. 

Additionally adding something to the `sqitch` command line app itself rather than the library would IMO be fine.

perigrin:
These test failures seem to be totally unrelated to my changes. I think `Dist::Zilla::Plugin::LocaleTextDomain` shifted beneath us and isn't installing clean. I had to force install it or one of it's dependencies to run tests before pushing this branch. I'm not sure how to move that ball forward here.

theory:
Yeah, I'll have to look in that.

In the meantime, I also recall that I didn't take the testing approach in the first place because when you load App::Sqitch, it doesn't load much; more gets loaded at runtime depending on the command, engine etc. So perhaps the test should search `lib` and load everything it can (ignoring things that die, since something can fail to load because, e.g., a DBI driver is missing), and *then* check for `Moose.pm` in `%INC`?

perigrin:
I thought about doing that but left it as is for now. I'll change the test to iterate through everything and check to make sure there is no Moose.

theory:
Okay, I think I have the CI issues [fixed](https://travis-ci.org/theory/sqitch/builds/236180948). Want to rebase?

perigrin:
Rebased and updated with some `File::Find` magic to test all the `.pm` files in `lib/`

perigrin:
It passes on everything but blead and that looks like a Travis error 😀

theory:
Perfect, thank you, @perigrin.

theory:
I disabled the blead build in d7a4634; seems to be a [bug in travis-perl](/travis-perl/helpers/issues/44).

PR #356, opened 2017-10-06
Fix grammar in Postgres Tutorial

avesus:
Its job is to test that the deploy did WAS it was supposed to -> Its job is to test that the deploy did WHAT it was supposed to.


theory:
Thank you!

PR #361, opened 2017-10-31
Cassandra support (very incomplete yet)

jviki:
Hello,

I am interested in including a support of the Cassandra engine into Sqitch. Here I am proposing an incomplete work. I am not a Perl programmer and nor an expert on databases and thus I did my best to write as much as I could.

Currently I am still unable to connect to the running Cassandra instance from Sqitch. I don't understand, how should I build the connection string. I could not find the right code that seems to be able to construct the `dbi:Cassandra:keyspace=...` like string for any other engine. The current runtime gives: 
```
Can't connect to data source 'dbname=testing' because I can't work out what driver to use (it doesn't seem to contain a 'dbi:driver:' prefix and the DBI_DRIVER env var is not set)
```
with configuration
```
[core]
  engine = cassandra
[engine "cassandra"]
  target = testing
[target "testing"]
  uri = db:cassandra:testing
```

I am also not sure how to implement the registry manipulations. The `App::Sqitch::Role::DBIEngine` requires some functions like `_ts_default`, etc. I could not find any documentation for this and I am not very sure what is expected there. Especially, the `_listagg_format`, `_no_table_error` and `_regex_op` are confusing for me. Moreover, there is no regex operator in Cassandra Query Language (probably, it can be implemented via some user-defined function).

I'd be glad if anybody can review the code or give some tips how to make it.

### DIFF
--- a/etc/templates/deploy/cassandra.tmpl
+++ b/etc/templates/deploy/cassandra.tmpl
@@ -1,0 +1,9 @@
+-- Deploy [% project %]:[% change %] to [% engine %]
+[% FOREACH item IN requires -%]
+-- requires: [% item %]
+[% END -%]
+[% FOREACH item IN conflicts -%]
+-- conflicts: [% item %]
+[% END -%]
+
+-- XXX Add DDLs here.
--- a/etc/templates/revert/cassandra.tmpl
+++ b/etc/templates/revert/cassandra.tmpl
@@ -1,0 +1,3 @@
+-- Revert [% project %]:[% change %] from [% engine %]
+
+-- XXX Add DDLs here.
--- a/etc/templates/verify/cassandra.tmpl
+++ b/etc/templates/verify/cassandra.tmpl
@@ -1,0 +1,3 @@
+-- Verify [% project %]:[% change %] on [% engine %]
+
+-- XXX Add verifications here.
--- a/lib/App/Sqitch/Command.pm
+++ b/lib/App/Sqitch/Command.pm
@@ -20,6 +20,7 @@ use constant ENGINES => qw(
     oracle
     firebird
     vertica
+    cassandra
 );
 
 has sqitch => (
--- a/lib/App/Sqitch/Engine/cassandra.cql
+++ b/lib/App/Sqitch/Engine/cassandra.cql
@@ -1,0 +1,157 @@
+CREATE KEYSPACE :"registry"
+WITH COMMENT = 'Sqitch database deployment metadata v1.0.',
+AND REPLICATION = {
+	'class': 'SimpleStrategy'
+	'replication_factor': 1
+},
+AND DURABLE_WRITES = true;
+
+CREATE TABLE :"registry".releases (
+	version         FLOAT PRIMARY KEY,
+	installed_at    TIMESTAMP.
+	installer_name  TEXT,
+	installer_email TEXT
+)
+WITH COMMENT = 'Sqitch registry releases.',
+AND CACHING = {
+	'keys': 'NONE',
+	'rows_per_partition': 'NONE'
+},
+AND GC_GRACE_SECONDS = 0,
+AND COMPRESSION = {
+	'class': 'LZ4Compressor'
+},
+AND CLUSTERING ORDER BY (version);
+
+CREATE TABLE :"registry".projects (
+	project       TEXT PRIMARY KEY,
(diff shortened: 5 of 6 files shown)


theory:
Hi @jviki, this is great to see, thank you! Let me take your questions one-at-a-time.

> I don't understand, how should I build the connection string.

There are two ways in which Sqitch connects to an engine: Via the command-line client (`csql` in your case, I think) and via an API. For the latter, all the existing engines are SQL-based, and so use the [DBI](https://metacpan.org/pod/DBI). I was assuming that Cassandra wouldn't use the DBI, but it turns out that [there is a driver](https://metacpan.org/module/DBD::Cassandra), so maybe it would work! The connection strings come from [URI::db](/theory/uri-db), which has a [Cassandra-specific subclass](/theory/uri-db/blob/master/lib/URI/cassandra.pm), much to my surprise, but it's certainly possible that it needs tweaking. For example, the [Cassandra DBI driver](https://metacpan.org/module/DBD::Cassandra) appears to use the attribute `db`, not `dbname`, which is the default for URI::db. So [URI::cassandra](/theory/uri-db/blob/master/lib/URI/cassandra.pm) probably needs a `dbi_param_map` method; see the one in [URI::sqlite](https://github.com/theory/uri-db/blob/master/lib/URI/mysql.pm#L9) for an example. It will end up something like:

```
sub _dbi_param_map {
    my $self = shift;
    return (
        [ hostname => scalar $self->host   ],
        [ port     => scalar $self->_port  ],
        [ database => scalar $self->dbname ],
    );
}
```

Or maybe it should use `keyspace` instead of `database`.

> The `App::Sqitch::Role::DBIEngine` requires some functions like `_ts_default`, etc. I could not find any documentation for this and I am not very sure what is expected there. Especially, the `_listagg_format`, `_no_table_error` and `_regex_op` are confusing for me. Moreover, there is no regex operator in Cassandra Query Language (probably, it can be implemented via some user-defined function).

Those are all pretty simple; look at the other engines to see what they should be.  For example, in the [Postgres engine](/theory/sqitch/blob/master/lib/App/Sqitch/Engine/pg.pm), you can see:

    sub _ts_default { 'clock_timestamp()' }

This is just a string to inject into a query for the current time. It should be as fine-grained as possible, since multiple changes can be deployed in less than a second. If that's not possible, you'll have to introduce a sleep of some kind, since the deploy time must be unique, and the order of changes is extremely important for Sqitch to work properly. You can see an example of such an approach in the [MySQL engine](/theory/sqitch/blob/master/lib/App/Sqitch/Engine/mysql.pm) (search for instances of `_fractional_seconds`; when that method returns false, the MySQL engine does more work to make sure things are ordered properly).

As for the others:

*   `_listagg_format`: String to inject into a query to aggregate all the values in a column over a series of rows. Should be in `sprintf` format. Here's the Postgres engine version:

        sub _listagg_format {
            q{ARRAY(SELECT * FROM UNNEST( array_agg(%s) ) a WHERE a IS NOT NULL)}
        }

    If csql doesn't support aggregates, it just means the you can't rely on the methods in DBIEngine that use it (`_select_state` and `_current_state`) and will have to implement them in your engine module. See the [Vertica engine](/theory/sqitch/blob/master/lib/App/Sqitch/Engine/vertica.pm) for an example of this pattern.

*   `_no_table_error`: Returns true if the last error from the database was for a missing table. This is used to detect when the registry doesn't exist. Here's the Postgres version:

        sub _no_table_error  {
            return $DBI::state && $DBI::state eq '42P01'; # undefined_table
        }

    I expect that you'll need to do some experiments with DBD::Cassandra and see what `$DBI::state` is set to after you try to query a keyspace that doesn't exist.

*   `_regex_op`: The database-native regular expression operator. This is used to search events. If Cassandra doesn't have a regex operator, perhaps it has some other kind of fuzzy matching operator? For example, the [Firebird engine](/theory/sqitch/blob/master/lib/App/Sqitch/Engine/firebird.pm) uses `SIMILAR TO`, which isn't quite a regex operator.

HTH!

theory:
Hi @jviki. Is this PR something you could see yourself returning too at some point?

jviki:
Sorry, initially I thought I would be able to make it but I am not interested in this topic anymore.

theory:
I get it, thanks for contributing what you had; may well help someone get a jump on an implementation later.

PR #374, opened 2018-03-05
Exit 0 for rebase and revert unless unsuccessful

### LINKED ISSUE #363: revert and rebase should exit with 0 except in case of error
Similar to the issue #147, the revert command returns 1 if there is nothing to revert, and 2 if there is an actual error with the reversion. Per the discussion at that issue, this is unexpected and maybe undesirable, as most applications expect 0 unless the operation actually failed.

The known workaround is to test the return code and only throw an error if the value is >= 2.

kwakwaversal:
Changed the exit value for rebase and revert from 1 to 0 when there is
no work to do. This is to match a shell's expectation of non-zero exit
statuses only when a command is unsuccessful. Nothing to do is
considered successful.

Closes #363.

### DIFF
--- a/Changes
+++ b/Changes
@@ -23,6 +23,10 @@ Revision history for Perl extension App::Sqitch
      - Added the Exasol engine. Thanks to Johan Wärlander for the PR (#362)!
      - Fixed and issue where URI::db needed to be explicitly loaded. Thanks to
        Hugh Esco for the report (#370)!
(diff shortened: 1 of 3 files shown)


kwakwaversal:
I think I've got all of the places for `rebase` and `revert`. I've left `exitval => 1` for prompts that are abandoned. My reasoning being that the original command was exited, so unsuccessful.

PR #375, opened 2018-03-06
Update `psql` options to use a conninfo string

kwakwaversal:
`psql` now uses a conninfo string (e.g., `psql "host=foo dbname=bar"`)
to honour connection parameter key words for pg: targets. It can now
take advantage of the connection service file using
`db:pg:///?service=$PGSERVICE` as well as other connection parameters.

### DIFF
--- a/Changes
+++ b/Changes
@@ -28,6 +28,10 @@ Revision history for Perl extension App::Sqitch
        exit statuses only when a command is unsuccessful, as well as the
        behavior of `deploy` as of v0.995. Nothing to do is considered
        successful (#363).
+     - Update `psql` options to use a conninfo string to honour connection
+       parameter key words for pg: targets. It can now take advantage of the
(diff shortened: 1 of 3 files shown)


kwakwaversal:
Not sure if this needs to be documented somewhere, or we should assume that if someone knows about [libpq connection](https://www.postgresql.org/docs/8.4/static/libpq-connect.html) parameters they will be able to figure it out for themselves.

theory:
Why would we need to document it? I don't think it currently documents the options it uses (`--dbname`, etc.). Only the use of `--set` is documented. It's Sqitch target URIs that users need to understand.

kwakwaversal:
Makes sense, just double checking. (I instinctively want to document something that I deem adds to the codebase, but in this instance it's probably expected to work this way so unnecessary.)

theory:
Looks good. I suggest adding a test with query params, too.

PR #378, opened 2018-04-04
Update Debian installation instructions

ilmari:
- Sqitch has been included in Debian since 8, no need to specify sid/testing/unstable
- Debian 9 ships Firebird 3.0, which has a separate client package (but keep the instructions for 8 and earlier)


theory:
Thanks!

PR #397, opened 2018-08-10
Add Snowflake Engine

theory:
[Snowflake](https://www.snowflake.net/) is data warehouse database that run top of cloud storage like Amazon S3, taking advantage of jillions of buckets to massively parallelize queries on large data sets. It's quite SQL compliant, though super slow for connection operations.

This PR adds support for Snowflake to Sqitch. Because of Snowflake's decent SQL comparability, the engine is relatively small, with few workarounds required to get it to work properly.

I consider this implementation complete with two exceptions:

1. A [DBD::ODBC bug](/perl5-dbi/DBD-ODBC/issues/8) prevents UTF-8 read from the database from being converted to Unicode. @mjegh has a fix, so I expect to bump up the minimum require DBD::ODBC once it has been released.

2. There is no CI integration for Snowflake. I have a request into the company for an account that could be used with our [Travis CI instance](https://travis-ci.com/sqitchers/sqitch/) to allow tests to run against a live Snowflake database. I will add that configuration once I have it.

I would appreciate a review. My Perl is getting a little rusty!


theory:
Thanks @ssoriche! I'm still trying to get access to a Snowflake instance for CI, but might not wait before merging…

PR #398, opened 2018-08-15
Add `./Build bundle` command.

theory:
It uses Menlo to install all prereqs into a single directory, then
installs Sqitch into that directory and adds an appropriate `use lib`
line to the Sqitch app installed there. It should be that Sqitch can
be run from that directory without any other Perl dependencies other
tha Perl itself.

The hope is to be able to more easily create simple bundles for
distribution in environments such as Homebrew, a Docker image, or an
OS-specific binary installer.

But I don't know if this is the best way to go about this. I am way behind on Perl installer best practices; maybe there's a better approach? For example, I notice that even though I'm parsing the `cpanfile` to require no test modules, modules such as Test::Simple get installed anyway, presumably because one or more dependencies end up requiring it. Is there any way around that?

Making @autarch a reviewer, since I think he's likely to be most knowledgable about this sort of thing. But feedback from all would be greatly appreciated.


README.md
```diff
@@ -84,6 +84,14 @@ To install Sqitch from a distribution download, type the following:
     ./Build test
     ./Build install
 
+To install Sqitch and all of its dependencies into a single directory named
+`sqtich_bundle`, type:
```
autarch: typo here.

README.md
```diff
@@ -84,6 +84,14 @@ To install Sqitch from a distribution download, type the following:
...
+
+    cpan Carton
+    ./Build bundle --install_base sqitch_bundle
+
+After which, Sqtich can be run from `./sqitch_bundle/bin/sqitch`.
```
autarch: Same typo.

autarch:
It'd be nice if we cvould use [fatpack](https://metacpan.org/release/App-FatPacker). This is what cpanm uses to produce a single script with all the necessary code.

But I don't think it could ever work when some of the prereqs involve XS, since that needs to be compiled on the machine where you install, not by us.

inc/Module/Build/Sqitch.pm
```diff
@@ -196,9 +197,9 @@ sub fix_shebang_line {
         my $FIXIN = IO::File->new($file) or die "Can't process '$file': $!";
         local $/ = "\n";
         chomp(my $line = <$FIXIN>);
-        next unless $line =~ s/^\s*\#!\s*//;     # Not a shbang file.
+        next unless $line =~ s/^\s*\#!\s*//;     # Not a shebang file.
 
-        my ($cmd, $arg) = (split(' ', $line, 2), '');
+        my ($cmd, $arg) = (split(' ', $line, 2), '');
```
autarch: Pointless whitespace added?

inc/Module/Build/Sqitch.pm
```diff
@@ -224,4 +225,57 @@ sub fix_shebang_line {
     return $self->SUPER::fix_shebang_line(@_);
 }
 
+sub ACTION_bundle {
+    my ($self, @params) = @_;
+    my $base = $self->install_base or die "No --install_base specified\n";
+    $self->_runtime_cpanfile;
```
autarch: I don't think this is needed. You can pass args to Carton like `--without develop --without test`.

inc/Module/Build/Sqitch.pm
```diff
@@ -224,4 +225,57 @@ sub fix_shebang_line {
...
+    open my $in, '<', 'cpanfile' or die "Cannot open cpanfile: $!\n";
+    open my $out, '>', 'cpanfile.run' or die "Cannot open cpanfile.run: $!\n";
+    while (<$in>) {
+        print {$out} $_ if /\Are(quire|commend)s\b/;
+    }
+    close $in;
+    close $out;
+}
+
+sub _add_carton_lib {
```
autarch: If the goal is to produce a single distributable tree then I think this can be greatly simplfiied. Just throw this in all the scripts:

```perl
use FindBin qw( $Bin );
use lib "$Bin/../lib";
```

Maybe this could even be added to the scripts in the repo rather than munging them here.

inc/Module/Build/Sqitch.pm
```diff
@@ -224,4 +225,57 @@ sub fix_shebang_line {
...
+            print {$libout} $_;
+            next unless /^\s*\#!\s*/; # Look for shebang line.
+            print {$libout} "use lib '$lib';\n", <$libin>;
+        }
+        close $libin;
+        close $libout;
+        rename("$file.new", $file)
+            or die "Can't rename $file.new to $file: $!";
+        $self->make_executable($file);
+        chmod oct(555), $file;
```
autarch: Aren't these 2 lines doing the same thing?

theory: Sadly no, the permissions were different without the `chmod`. I can remove the `make_executable`, I think.

theory:
Appreciate the review, @autarch. I would think Fatpack might work, because the resulting "binary" would only be distributed to platforms that support it (Homebrew, macOS installer, Docker image, etc.). But that's a guess; I've never used it.

inc/Module/Build/Sqitch.pm
theory: Yeah, `t/sqitch` does that.

theory:
Mind having a look again, @autarch? I switched to Menlo (which is derived from cpanm) because a subclass lets me force it to install only runtime requirements. I've also added support for CPAN::META::Spec v2 optional features to control modules installed for various database engines. I revamped the Homebrew tap to [a single formula](https://github.com/sqitchers/homebrew-sqitch/blob/bundle/Formula/sqitch.rb) to demonstrate that it works. I'm pretty happy with it, but would greatly appreciate one or more critical reviews.

PR #409, opened 2018-10-06
Fix the "conmand" typo in the docs for sqitch upgrade

AnianZ:
In the error message for an outdated sqitch registry the user is prompted to run the "conmand" `sqitch upgrade`. This typo occurred on multiple locations. I guess this is just a copy & pasted typo so I fixed it.


AnianZ:
I also noticed there is a file for `de` as a language but the content is English. Would it help to translate that file?

PR #413, opened 2018-12-18
Quote mysql.exe calls on Windows correctly because IPC::S::S doesn't.

### LINKED ISSUE #262: Sqitch on Windows 10 issue
Hi,

I'm trying to set up Sqitch to work with a localhost MySQL server on my Windows 10 machine. I installed Sqitch through ActivePerl with no issues and I have WAMP set up on my pc. Both are on the same hard drive and in the same directory. MySQL 5.6.17 is what I have currently installed.

 However, when I run `sqitch deploy` from my project's db folder it says 
`adding the registry tables to db:mysql://root:@localhost/sqitch` 
then immediately displays 
`"mysql.exe" failed to start: "The system cannot find the file specified"`

I looked and mysql.exe is definitely in my WAMP directory and I can access my localhost db directly from WAMP without issue so I'm pretty stumped as to what the issue can be.

wchristian:
run() here calls IPC::S::S' runx(), which has two paths. On non-windows systems it calls `system { $command } $command, @args` which works just fine even with complex stuff in the args. On Windows however it calls `_spawn_or_die($command, "$command @args")`, which smashes all the arguments together separated by spaces without *any* additional quoting. This results in a garbage process call which mysql consequently rewards with RTFM.

This fix quotes the arguments on windows, and takes care to leave the executable name separate to allow proper naming of the created process object.

fixes #262

### DIFF
--- a/lib/App/Sqitch.pm
+++ b/lib/App/Sqitch.pm
@@ -358,7 +358,7 @@ sub run {
         ( my $msg = shift ) =~ s/\s+at\s+.+/\n/ms;
         die $msg;
     };
-    runx @_;
+    runx $^O eq 'MSWin32' ? ( shift, $self->quote_shell(@_) ) : @_;
     return $self;
 }
 


theory:
I'm torn about this. I intentionally didn't make this change because I was optimistic that pjf/ipc-system-simple#29 would get merged and released. Perhaps too optimistic. :-( Perhaps we should try to figure proof it to only quote the words if the version is `<= 1.25`.

wchristian:
I have absolutely no preference for this at the moment. sqitch is simply broken on windows. Unusuable. No good. Any action is better than no action.

theory:
Okay, done in 0148359b4fef572f8b048199ea745c45ae684ee1.

wchristian:
Looks excellent! 👍

theory:
FYI, 0148359 had a bug. I added windows testing and found it. Fixed in c697c9b.

Issue #275, opened 2015-12-03
Add --set option to engine, target


theory:
This will add a `variables` section to the specific engine or target, complementing the variables that can be set by the `deploy` command. Don't think we need `--set` for init.

nugget:
This would be a really powerful addition.  Is this still on the roadmap?

theory:
Yep, but I don't have much time to hack on Sqitch these days. Want to take a stab at it?

nugget:
I'd love to, but my Perl is weak and I can't even get Dist:Zilla to install.  Seems unapproachable to me, sorry.

kwakwaversal:
I'll take a look at doing this. How do you envisage the `variables` section to look in the `sqitch.conf`?

```ini
[core]
        engine = pg
[engine "pg"]
        target = db:pg:
        [variables]
                ON_ERROR_STOP = 1
                foo = baz
[target "dev"]
        uri = db:pg://username:password@localhost/somedatabase
        [variables]
                ON_ERROR_STOP = 1
                foo = baz
```

I'm just guessing here. I've not checked the code for parsing the config to see if it supports multi-dimensional config.

nugget:
In my googling earlier today I found references to a syntax like `set        = foo=bar` ([ref](http://search.cpan.org/~dwheeler/App-Sqitch-0.992/lib/sqitch-config.pod)) from older versions of Sqitch. 

Is that approach deprecated now in favor of the `[verify "variables"]` style blocks (which aren't directly usable as @kwakwaversal points out above?

Both the nested section and the `set = foo= bar` seem ungainly, but I'm not coming up with a better approach.

theory:
[...]

PR #417, opened 2018-12-30
Target variables

theory:
Add support for core, engine, and target variable configuration, as requested in #275. The simplest way to use them is via the `--set` option on the `init`, `engine`, and `target` commands. These allow the configuration of database client variables for specific engines and targets, as well as defaults that apply to all change execution commands (`deploy`, `revert`, `verify`, `rebase`, and `checkout`). The commands merge the variables from each level in this priority order:

* `--set-deploy` and `--set-revert` options on `rebase` and `checkout`.
* `--set` option
* `target.$target.variables`
* `engine.$engine.variables`
* `deploy.variables`, `revert.variables`, and `verify.variables`
* `core.variables`

See `sqitch-configuration` for general documentation of of the hierarchy for merging variables and the documentation for each command for specifics.


theory:
Thanks @ssoriche. I'll be rebasing before I merge.

Issue #250, opened 2015-08-13
Move All Non-Meta Options to Commands?


theory:
There are a few options left in the core that might be better moved to the commands that require them:

```
    plan-file|f=s
    engine=s
    registry=s
    client|db-client=s
    db-name|d=s
    db-username|db-user|u=s
    db-host|h=s
    db-port|p=i
    top-dir|dir=s
```

This can be done by creating roles that inject sets of options (like TargetConfigCommand does), and then add them to the commands that need them. The maybe then inject them into `$sqitch->options` so that they are available everywhere?

After that, we can probably remove the requirement that core and command options be separate: let them appear anywhere. Will need to merge those that overlap (I think it's just `--verbose` in a few commands).

theory:
The work done for TargetConfigCommand in 418423b71 did some of this already. Notably, `--engine` was added to `init`, and probably isn't needed anywhere else; that's what the `engine` command is for. It isn't even needed for `add`, because it can be passed as a target name:

    sqitch add foo -m foo db:pg:

Or even just the engine:

    sqitch add foo -m foo pg

418423b71 also added the `--top-dir` and `--dir` options to the `init`, `target`, and `engine` commands. I think it makes a ton of sense to let those options set configurations rather than be used on the command-line. I'm a little torn on `--top-dir`, since I've often used it myself, especially in projects where there are several plans that rely on the same configuration and same engines & targets. So maybe that one's best left as a core option.

I'm thinking that `--plan-file` should stay a core option. Not *all* commands use it (`config` doesn't, for example), but the plan file is so central to everything Sqitch does, I think it needs to stay.

The other options need to be added to the following commands:

* `registry`:
    + `deploy`
    + `revert`
    + `verify`
    + `checkout`
    + `rebase`
    + `log`
    + `show`
    + `status`
    + `upgrade`

PR #421, opened 2019-01-11
Option unification

theory:
This pull request changes the way Sqitch handles command line argument parsing to allow both core and command options to appear either before or after the command.  Previously, core options had to come before the command name, and command options after. No more. The caveat  is that command options that take arguments should either appear after  the command or use the `--opt=val` syntax instead of `--opt val`, so  that Sqitch doesn't think `val` is the command. Even in that case, it  will keep searching the rest of the arguments to find a valid command. However, to minimize this challenge, the documentation now suggests  and demonstrates putting all options after the command, like so: 

    sqitch [command] [options]

With the ability to mix and match core and command options, in order to eliminate conflicts between the two categories of options, this PR also moves all non-global options from the core to the commands that need them (#250). Only options that affect output and runtime context remain in the core:

* -C --chdir --cd <dir>  Change to directory before performing any actions
*    --etc-path          Print the path to the etc directory and exit
*    --no-pager          Do not pipe output into a pager
*    --quiet             Quiet mode with non-error output suppressed
* -V --verbose           Increment verbosity
*    --version           Print the version number and exit
*    --help              Show a list of commands and exit
*    --man               Print the introductory documentation and exit

Note that single-character aliases are now uppercase for core options (`-v` is still supported for `--verbose`, but no longer documented). The options that were moved include:

-   plan-file|f=s
-   engine=s
-   registry=s
-   client|db-client=s
-   db-name|d=s
-   db-username|db-user|u=s
-   db-host|h=s
-   db-port|p=i
-   top-dir|dir=s

I believe all the commands that need to use one or more of these options now support them. The one exception I can think of is the `plan` command, which does not have the `--plan-file` option. I think this is okay because you can simply pass the plan file path to the command:

    sqitch plan sql/widgets.plan

This PR also deprecates (with a warning) the `--top-dir` option from all but the configuration commands (`init`, `engine`, `target`) because the new `--chdir` option serves just as well. Consider a project like so:

```
sqitch.conf
db/sqitch.conf
db/go.plan
db/deploy/foo.sql
db/revert/foo.sql
db/verify/foo.sql
```

Traditionally, one could deploy this project like so:

```
sqitch --top-dir --plan-file db/go.plan db deploy db:pg:mydb
```

But that doesn't read `db/sqitch.conf`. This variant, however, which changes into the `db` directory before reading any files or executing any commands, reads both configuration files:

```
sqitch deploy --chdir db --plan-file go.plan db:pg:mydb
```

Because Sqitch's configuration file API reads all instances of `sqitch.conf` from parent directories in its path.

The deprecation warning suggest using `--chdir` instead; the `--top-dir` option will be removed in the future.

There are a few other potentially breaking changes in this PR, though I think they're all edge cases:

* The `--verbose` option on the `--engine` and `--target` commands has been removed, but no visible change should be apparent, since those commands now read the core `--verbose` option.
* The undocumented `--dir` alias for `--top-dir` has been removed, as it conflicted with the option of the same name but different meaning in the `init`, `engine`, and `target` commands.
* The `-d` alias for `--set-deploy` in the `rebase` and `checkout` commands has been changed to `-e` so as not to conflict with the `-d` alias for `--db-name`.
* Removed the `--engine` option from all but the `init` command (which is the only place the tutorials and other docs ever used it). Other commands are able to parse the engine name as an argument, e.g., `sqitch deploy mysql`, or implicitly as part of a target, as in `sqitch revert db:pg:tryme'.

This PR also removes these core options, which long ago started issuing warnings:

-   deploy-dir=s
-   revert-dir=s
-   verify-dir|test-dir=s

It also removes the reading of old-style `core.$engine_name` configurations, which also have long triggered a warning and a suggestion to run `sqitch engine upgrade-config` to update them. That action is still around for any stragglers.


theory:
Went ahead and deprecated `--top-dir` in 
200a443..37ec0710.

theory:
Merged in c0be46a49c80b1ffc06bf65fbe7470e0bbca091f.

PR #433, opened 2019-02-26
remove the use of File::HomeDir

karenetheridge:
..as this pulls in extra dependencies on darwin, which complicates static deployment files.
Also removed (nearby) dangerous uses of "my ... = ... or ..." which can result in inadvertent closures.


theory:
Well, is there not a better module for this, then?

theory:
Also, can you change it to merge into the develop branch?

karenetheridge:
I have rebased against develop. I would suggest you update the settings for the repository so that clones automatically base from develop rather than master, to avoid future confusion.

There is File::HomeDir::Tiny, but it has an outstanding bug, resulting in windows malfunctions: https://rt.cpan.org/Ticket/Display.html?id=127425

theory:
Yeah, @ssoriche and I tweaked the branch settings this afternoon. Lesson learned.

lib/App/Sqitch/Config.pm
```diff
@@ -21,10 +21,8 @@ has '+encoding' => ( default => 'UTF-8' );
 my $SYSTEM_DIR = undef;
 
 sub user_dir {
-    require File::HomeDir;
-    my $hd = File::HomeDir->my_home or hurl config => __(
-        "Could not determine home directory"
-    );
+    my $hd = $^O eq 'MSWin32' && "$]" < '5.016' ? $ENV{HOME} || $ENV{USERPROFILE} : (glob('~'))[0];
```
theory: Should probably use App::Sqitch::ISWIN where appropriate.

karenetheridge:
Thanks!

PR #435, opened 2019-02-27
Replace "sane" with "sensible".

theory:
Using a mental health state as a signifier of something good (sane) or bad
(insane) falls in the category of being insensitive to sufferers of mental
illness. So lets go with "sensible", which captures the same general idea.


theory:
I can't change the name of the presentation from a few years ago; perhaps there are more recent presentations we can point to, instead?

PR #446, opened 2019-04-04
Switch to dotted decimal version strings

theory:
The use of SemVer was likely to be troublesome for modules declaring dependencies, so drop it in favor of three-part dotted decimal versions in strings, with a preceding v. Should work well on Perl 5.10 and later, which is all Sqitch supports. See [this blog post](http://blogs.perl.org/users/grinnz/2018/04/a-guide-to-versions-in-perl.html) for more details:

Also, automate version setting by using [Dist::Zilla::Plugin::OurPkgVersion](https://metacpan.org/pod/Dist::Zilla::Plugin::OurPkgVersion). This lets us set the version in one place, the dist.ini (well, and also `dist/sqitch.spec`), and Dist::Zilla will add the version to all the files.

For prereleases, we should mark versions for prerelease by using the `--trial` option when making a release:

    dzil release --trial

Furthermore, once v1.0.0 goes out, we should increment develop to 1.0.0.1 --- that is, always append a fourth part in the develop branch, so it's always higher than the current release (in master), but lower than the next release will be, when we should drop the fourth part and increment the third or second parts. That is, to merge a release into master, make its version 1.0.1 or 1.1.0. Then revise develop to be 1.0.1.1 or 1.1.0.1.


theory:
Oh, and I edited Changes and tweaked doc wording in a few places, too.

PR #456, opened 2019-05-01
Added missing German translation

tiguchi:
Also changed a translation that sounded a bit off.

@theory You mentioned in my last PR that there are [2 additional translations](https://github.com/sqitchers/sqitch/pull/451#issuecomment-485821346):  I could only find one 😕 Let me know if there's another one, then I'll update this PR


theory:
Thank you!

Issue #460, opened 2019-05-19
sqlite3 first time deploy doesn't work if ~/.sqliterc exists


kyzn:
Hi! Thanks for maintaining sqitch, it's a great library. I hope you find this issue (and proposed solution to it) helpful. Here goes.

I've recently tried to deploy a sqitch for the first time in a library (so no sqitch.db exists yet). It failed with an interesting error message:

```
Adding registry tables to db:sqlite:sqitch.db
-- Loading resources from /home/kivanc/.sqliterc
Use of uninitialized value in split at /home/kivanc/perl5/perlbrew/perls/perl-5.28.1/lib/site_perl/5.28.1/App/Sqitch/Engine/sqlite.pm line 117.
Use of uninitialized value $v[0] in numeric gt (>) at /home/kivanc/perl5/perlbrew/perls/perl-5.28.1/lib/site_perl/5.28.1/App/Sqitch/Engine/sqlite.pm line 120.
Use of uninitialized value $v[0] in numeric eq (==) at /home/kivanc/perl5/perlbrew/perls/perl-5.28.1/lib/site_perl/5.28.1/App/Sqitch/Engine/sqlite.pm line 120.
Sqitch requires SQLite 3.3.9 or later; sqlite3 is
```

Now, it looks like it can't get my sqlite3 version, which happens with running `sqlite3 -version` through `IPC::System::Simple`'s `capturex`. Also note the message about `Loading resources`.

I figured what happens is related to those two. If I don't have a `.sqliterc` file in my home directory, this is the output from `sqlite3 -version`:

```
0> use IPC::System::Simple qw/capturex/
1> capturex('sqlite3','-version');
$res[0] = '3.11.0 2016-02-15 17:29:24 3d862f207e3adc00f78066799ac5a8c282430a5f
'
```

And if I do have it, here's what happens:
```
2> capturex('sqlite3','-version');
-- Loading resources from /home/kivanc/.sqliterc
$res[1] = [
  '
',
  '3.11.0 2016-02-15 17:29:24 3d862f207e3adc00f78066799ac5a8c282430a5f
'
]
```

What happened here? There's now some additional output that went to STDOUT. And capturex returned an array with two items, first one being a `\n`.

The way `App::Sqitch::Engine::sqlite` checks for installed sqlite3 version is this: It calls `App::Sqitch`'s `probe`, which is `App::Sqitch`'s `capture` + a `chomp`. `capture` calls `capturex` (from above) and returns whatever it returns.

[...]

PR #461, opened 2019-05-19
Enforce capture to return scalar value

kyzn:
Otherwise probe might be missing parts of results
See https://github.com/sqitchers/sqitch/issues/460


kyzn:
Cancelling due comments in #460. I'll submit a separate patch soon.

PR #465, opened 2019-05-27
Enforce scalar context when parsing sqlite version

kyzn:
sqlite3 outputs an extra newline when v3.11.0 is installed
and ~/.sqliterc exists. This prevents us from parsing version.

Using capture and enforcing scalar context fixes the issue.
This commit addresses GitHub issue #460.

### DIFF
--- a/lib/App/Sqitch/Engine/sqlite.pm
+++ b/lib/App/Sqitch/Engine/sqlite.pm
@@ -115,7 +115,7 @@ has _sqlite3 => (
 
         # Make sure we can use this version of SQLite.
         my @v = split /[.]/ => (
-            split / / => $self->sqitch->probe( $self->client, '-version' )
+            split / / => scalar $self->sqitch->capture( $self->client, '-version' )
         )[0];
         hurl sqlite => __x(
             'Sqitch requires SQLite 3.3.9 or later; {client} is {version}',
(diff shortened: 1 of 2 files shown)


theory:
Awesome. Could you also add a test case that matches the bug? That is, set up `$sqlite_version` with the value you saw originally, with the extra newline, to make sure it works and no one tries to change it back in the future?

kyzn:
Sure! Give me a minute.

kyzn:
Oh and I found another issue in updated test suite. When I re-mocked `capture` I forgot to restore `sqlite_version` of `3.7` that was declared on line 57. This was causing rest of test to think capture returns `1.0.0`. Fixing that as well.

kyzn:
So here's new changes:
- In first commit, I have made sure to reset `sqlite_version` back to `3.7.12` so the rest of tests can run fine.
- Second commit should not have any functional affects, It's more of a "rearrange" so that third commit is easier to read. I had to drop `qw` in order to test version `\n3.11.0`. I also created a new variable named `$v_print`: This is so that if you are running `prove -v` one line of test is not spread across two lines.
- Third commit adds the new tests.

Let me know if you would like to update any pieces 🙂

kyzn:
Looking at failing windows test..

Issue #475, opened 2019-06-07
Add command to check list deployed changes for which the deploy script has changed


matthieu-foucault:
I'm relatively new to using sqitch so let me know if there's something obvious that I'm missing here.

Here is a scenario describing the behavior I would expect from `sqitch`:
 1. deploy a plan containg one or more changes
 1. make changes to some deploy scripts
 1. `sqitch verify` should output which scripts should be re-deployed, by comparing the sha1 sum of the scripts in the working directory to the ones in the `changes` table

This would be particularly useful when developing deploy scripts. AFAIK, there is no easy way to tell which scripts need to be redeployed when switching branches. One would expect `sqitch checkout` to ensure that the database has a state that matches the deploy scripts in the checked out branch, but that does not seem to be the case.

I'm going to start working on implementing this, and will submit a PR if you think that this feature should be added.

ssoriche:
Welcome to sqitch!

Sqitch changes are deployed in order. If you need to go back and change a deploy script that has already been deployed, that’s when `sqitch rework` is used. The other option is creating a new deploy script that updates to the way you want the database to appear now.

If you’re working in your development environment and you’re making changes to the sqitch files to make sure they work, you can use `sqitch revert` to undo the changes that have been applied and then redeploy them. The `sqitch revert` command also takes a change name parameter to revert to, allowing changes to be reverted to a previous change.

I hope this helps clarify the workings for you.

matthieu-foucault:
We're working in our development environment. We are starting to have a significant number of scripts, and we want to improve our development process. 

The problem with `sqitch revert` is figuring out which change to revert to. This is something that needs to be done multiple times a day, as we switch branches, pull new changes, etc. The needed commands will be implemented in git hooks eventually, but I feel that something is missing here, as I haven't found a sqitch command indicating the developer that their working directory contains deploy scripts what do not match what is deployed.

When deploying, sqitch stores the SHA1 of the deploy script, so it would be easy, when doing a `sqitch verify` (or whatever command is most appropriate), to check which deploy scripts have changed in the working directory.

theory:
Good idea. Not for `sqitch verify`, but a new command, "Show me what's changed" sort of thing.

matthieu-foucault:
Any preference on how that command should be named?
I asked around and we have those for now:
 - `sqitch check-workdir`
 - `sqitch checksum`
 - `sqitch digest`

ssoriche:
[...]

PR #477, opened 2019-06-10
Add a sqitch check command that prints changes in plan for which script has changed

matthieu-foucault:
Re #475 

I'm thinking about adding a `--redeploy` option that would revert to the appropriate change and deploy the plan again, so that the database has deployed changes matching the working directoy.

It's my first time writing in Perl, so let me know if I am missing some language idioms, or if there are some lines in there that look horribly wrong.

It's still a draft. I haven't written any test yet, and I am sure I am missing a lot of edge cases.
Happy to receive feedback either way :)

### DIFF
--- a/lib/App/Sqitch/Command/check.pm
+++ b/lib/App/Sqitch/Command/check.pm
@@ -164,10 +164,11 @@ The Sqitch command-line client.
 =head1 Author
 
 David E. Wheeler <david@justatheory.com>
+Matthieu Foucault <matthieu@button.is>
 
 =head1 License
 
-Copyright (c) 2012-2018 iovation Inc.
+Copyright (c) 2012-2019 iovation Inc., Button Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/lib/App/Sqitch/Command/check.pm
+++ b/lib/App/Sqitch/Command/check.pm
@@ -4,11 +4,11 @@ use 5.010;
 use strict;
 use warnings;
 use utf8;
-use Locale::TextDomain qw(App-Sqitch);
-use App::Sqitch::X qw(hurl);
 use Moo;
-use App::Sqitch::Types qw(Str Target Engine Change);
-use Try::Tiny;
+use Types::Standard qw(Str HashRef);
+use App::Sqitch::X qw(hurl);
+use Locale::TextDomain qw(App-Sqitch);
+use List::Util qw(first);
 use namespace::autoclean;
 
 extends 'App::Sqitch::Command';
@@ -17,62 +17,69 @@ with 'App::Sqitch::Role::ConnectingCommand';
 
 # VERSION
 
-has target_name => (
+has target => (
     is  => 'ro',
     isa => Str,
 );
 
-has target => (
-    is      => 'rw',
-    isa     => Target,
-    handles => [qw(engine plan plan_file)],
+has from_change => (
+    is  => 'ro',
+    isa => Str,
 );
 
-has date_format => (
-    is      => 'ro',
-    lazy    => 1,
-    isa     => Str,
-    default => sub {
-        shift->sqitch->config->get( key => 'status.date_format' ) || 'iso'
-    }
+has to_change => (
+    is  => 'ro',
+    isa => Str,
 );
 
-has project => (
-    is      => 'ro',
-    isa     => Str,
-    lazy    => 1,
-    default => sub {
-        my $self = shift;
-        try { $self->plan->project } catch {
-            # Just die on parse and I/O errors.
-            die $_ if try { $_->ident eq 'parse' || $_->ident eq 'io' };
-
-            # Try to extract a project name from the registry.
-            my $engine = $self->engine;
-            hurl status => __ 'Database not initialized for Sqitch'
-                unless $engine->initialized;
-            my @projs = $engine->registered_projects
-                or hurl status => __ 'No projects registered';
-            hurl status => __x(
-                'Use --project to select which project to query: {projects}',
-                projects => join __ ', ', @projs,
-            ) if @projs > 1;
-            return $projs[0];
-        };
-    },
+has variables => (
+    is       => 'ro',
+    isa      => HashRef,
+    lazy     => 1,
+    default  => sub { {} },
 );
 
 sub options {
     return qw(
-        project=s
         target|t=s
+        from-change|from=s
+        to-change|to=s
+        set|s=s%
+    );
+}
+
+sub configure {
+    my ( $class, $config, $opt ) = @_;
+
+    my %params = map {
+        $_ => $opt->{$_}
+    } grep {
+        exists $opt->{$_}
+    } qw(target from_change to_change);
+
+    if ( my $vars = $opt->{set} ) {
+        $params{variables} = $vars;
+    }
+
+    return \%params;
+}
+
+sub _collect_vars {
+    my ($self, $target) = @_;
+    my $cfg = $self->sqitch->config;
+    return (
+        %{ $cfg->get_section(section => 'core.variables') },
+        %{ $cfg->get_section(section => 'deploy.variables') },
+        %{ $cfg->get_section(section => 'verify.variables') },
+        %{ $target->variables }, # includes engine
+        %{ $self->variables },   # --set
     );
 }
 
 sub execute {
     my $self = shift;
-    my ($targets) = $self->parse_args(
-        target => $self->target_name,
+    my ($targets, $changes) = $self->parse_args(
+        target => $self->target,
         args   => \@_,
     );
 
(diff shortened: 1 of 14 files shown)


theory:
At a glance, it looks pretty complete, thanks! It will likely be a week or so before I can look at it in detail due to a super busy work week.

> I'm thinking about adding a `--redeploy` option that would revert to the appropriate change and deploy the plan again, so that the database has deployed changes matching the working directoy.

Sounds like the [`rebase` command](https://sqitch.org/docs/manual/sqitch-rebase/). Or do I misunderstand? Oh, maybe you mean so it knows what version to revert to? Might be useful to add some sort of `--revised` option to the `rebase` command, I think.

matthieu-foucault:
It's the `rebase` command indeed. I can add a `--revised` option that finds the parent of the first script that is different in the working directory and rebases onto that

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+    isa => Str,
+);
+
+has target => (
+    is      => 'rw',
+    isa     => Target,
+    handles => [qw(engine plan plan_file)],
+);
+
+has date_format => (
```
theory: Don't think you need this attribute. Should probably model attributes based on the `verify` command.

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+    is      => 'ro',
+    lazy    => 1,
+    isa     => Str,
+    default => sub {
+        shift->sqitch->config->get( key => 'status.date_format' ) || 'iso'
+    }
+);
+
+has project => (
```
theory: Don't think you nee the `project` attribute, either, since part of the point of `check` is to compare the registry to the plan, so you need a plan file which includes the project. Honestly, I think the only attributes you need are `target` and `variables` (copy from `verify`).

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+                projects => join __ ', ', @projs,
+            ) if @projs > 1;
+            return $projs[0];
+        };
+    },
+);
+
+sub options {
+    return qw(
+        project=s
```
theory: Remove `project` and add `set|s=s%` for variables.

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+        target|t=s
+    );
+}
+
```
theory: Add the `configure` method to merge variables once they're added.

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+    # Warn on multiple targets.
+    my $target = shift @{ $targets };
+    $self->warn(__x(
+        'Too many targets specified; connecting to {target}',
+        target => $target->name,
+    )) if @{ $targets };
+
+    # Good to go.
+    $self->target($target);
+    my $engine = $target->engine;
```
theory: To follow the other database commands, I'd move most of the rest of this method to a `check` method in App::Sqitch::Engine. I think it's likely to share quite a bit of its logic with the `verify` method in that class, so it might make sense to factor out the common bits.

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+
+    # Where are we?
+    $self->comment( __x 'On database {db}', db => $engine->destination );
+
+    # Exit with status 1 on no state, probably not expected.
+    my $state = try {
```
theory: This doesn't seem necessary, since you never use the `$state` in the remainder of the method.

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+        $deployed_script_hashes{$change->{'id'}} = $change->{'script_hash'};
+    }
+
+    my $workdir_plan = $target->plan;
+    my @workdir_changes = $workdir_plan->changes;
+    foreach my $change (@workdir_changes) {
+        $self->comment(__x(
+            'Working directory script {script_file} is different from deployed script',
+            script_file => $change->deploy_file
+        )) if $change->script_hash ne $deployed_script_hashes{$change->id};
```
theory: We should give some thought to the output of failed checks. Should they be parsable? Have some kind of header or something? Also, once you find one bad ID, all the remaining IDs will be bad, so might as well say something like `"Script signatures diverge at change {change}"` and stop processing. In fact, you should be able to just iterate over the plan and the deployed changes without using a lookup hash, because the first time something is out of order it's all over.

Once you find something that's not right, set up a variable to exit with a non-zero status, as well. I think you can just set a boolean, and then raise an error using `hurl`; maybe have the exit value be the number of failures (just going to be one for now, but as more checks are added, it can go up), something like:

``` perl
        hurl {
            ident => 'check',
            message => __np(
                'Failed one check',
                'Failed {count} checks',
                $num_failed,
                count => $num_failed,
            ),
            exitval => $num_failed,
        } if $num_failed;
```

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+
+=item L<sqitch>
+
+The Sqitch command-line client.
+
+=back
+
+=head1 Author
+
+David E. Wheeler <david@justatheory.com>
```
theory: Give yourself some credit!

lib/App/Sqitch/Command/check.pm
```diff
@@ -0,0 +1,208 @@
...
+
+=head1 License
+
+Copyright (c) 2012-2018 iovation Inc.
```
theory: You can include yourself (or your org) on this line along with iovation.

lib/App/Sqitch/Role/DBIEngine.pm
```diff
@@ -756,7 +756,7 @@ sub deployed_changes {
     } @{ $self->dbh->selectall_arrayref(qq{
         SELECT c.change_id AS id, c.change AS name, c.project, c.note,
                $tscol AS "timestamp", c.planner_name, c.planner_email,
-               $tagcol AS tags
+               $tagcol AS tags, c.script_hash
```
theory: The addition of `script_hash` makes a lot of sense here, but you almost certainly need to update unit tests for these methods, especially in `t/engine.t` and `t/lib/DBIEngineTest.pm`. Also check the specific engines (subclasses of App::Sqitch::Engine) to see if any override these methods, as those overrides will need to be modified, too.

lib/sqitch-check.pod
```diff
@@ -0,0 +1,113 @@
+=head1 Name
+
+sqitch-check - Runs checks
```
theory: I'd make this a little more verbose, something like `Checks for divergences between planned and deployed changes`.

t/lib/DBIEngineTest.pm
```diff
@@ -799,6 +799,7 @@ sub run {
             planner_name  => $change->planner_name,
             planner_email => $change->planner_email,
             tags          => ['@alpha'],
+            script_hash   => $change->script_hash,
```
theory: Oh, you did update these tests. Sweet!

matthieu-foucault:
Thanks for the review @theory ! I'll have time to work on it this week, and add the `revert --revised` option mentioned above.

matthieu-foucault:
Both `check` and `rebase --revised` work at this point. I should be able to work on adding tests tomorrow.

matthieu-foucault:
@theory There are still tests running on TavisCI, but I think this is good for a review :tada:

matthieu-foucault:
It seems that the windows build can't find `Test::MockObject::Extends`, which I added to be able to mock a `script_hash` change. @theory any idea on how to fix that?

theory:
Thanks for that update. Reviewing this PR is on my list, but not sure when I will get to it. Life pretty bananas right now…appreciate your patience.

lib/App/Sqitch/Command/rebase.pm
```diff
@@ -27,10 +27,17 @@ has upto_change => (
...
+    is  => 'ro',
+    isa => Bool,
+    default => 0,
+);
+
 sub options {
     return qw(
         onto-change|onto=s
         upto-change|upto=s
+        revised!
```
theory: I don't think we want this to be negate-able; the `!` means that there is also a `--no-revised` option, which doesn't seem necessary. Am I missing something?

lib/App/Sqitch/Engine.pm
```diff
@@ -1094,6 +1106,80 @@ sub upgrade_registry {
...
+
+    foreach my $change (@deployed_changes) {
+        $i++;
+        return $i if $i + $from_idx >= $plan->count || $change->script_hash ne $plan->change_at($i + $from_idx)->script_hash;
+    }
+
+    return -1;
+}
+
+sub planned_deployed_common_ancestor_id {
```
theory: I didn't see a test for this method, or for `_find_planned_deployed_divergence_idx`. Are they implicit in the tests for `check`?

lib/sqitch-rebase.pod
```diff
@@ -63,6 +65,13 @@ L<sqitchchanges> for the various ways in which changes can be specified.
 Specify the deployment change. Defaults to the last point in the plan. See
 L<sqitchchanges> for the various ways in which changes can be specified.
 
+=item C<--revised>
+
+Finds the change to revert onto based on the output of L<sqitch-check>. The
```
theory: Not sure this is accurate: `based on the output of L<sqitch-check>`. Is there output from `sqitch check` when you run this? I thought it was simply that it would rebase on the last change that does not conflict between the plan and the deployed changes. I think you can omit the part about the output of sqitch check from this description.

lib/sqitch-check.pod
```diff
@@ -0,0 +1,113 @@
...
+
+sqitch-check - Checks for divergences between planned and deployed changes.
+
+=head1 Synopsis
+
+  sqitch check [options] [<database>]
+
+=head1 Description
+
+Runs various checks on the state of the working directory and the database.
```
theory: Is it "various checks" at this point? Seems like there's just the one at the moment; perhaps mention that more may be added in the future?

theory:
Hey @matthieu-foucault, would you give me permission to push to your branch? Details [here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork). I've rebased it and made a few tweaks. If I can push to your branch, it will keep this PR in-tact, which is nice for history, and I can get it merged post-haste. Thanks!

matthieu-foucault:
Hi @theory , thanks for taking over that PR. 
Unfortunately, I cannot allow edits from maintainers as the PR is coming from an org repo, not a my personal user repo (see https://github.community/t5/How-to-use-Git-and-GitHub/How-can-we-enable-allow-edits-from-maintainers-by-default/m-p/51971/highlight/true#M11129)

theory:
Bummer. Guess I'll have to close this PR and make a new one. :-(

PR #485, opened 2019-08-27
Only add password to the mysql command when no .my.cnf is used

kielstr:
Later versions of MySQL generate the following warning when using the --password command-line option.

```
mysql: [Warning] Using a password on the command line interface can be insecure.
```

Very annoying. This change avoids it when using a .my.cnf file.

### DIFF
--- a/lib/App/Sqitch/Engine/mysql.pm
+++ b/lib/App/Sqitch/Engine/mysql.pm
@@ -171,7 +171,8 @@ has _mysql => (
 
         # Special-case --password, which requires = before the value. O_o
         if (my $pw = $self->password) {
-            push @ret, "--password=$pw";
+            push @ret, "--password=$pw"
+               unless $self->_mycnf->{password};
         }
 
         # Options to keep things quiet.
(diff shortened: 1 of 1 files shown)


lib/App/Sqitch/Engine/mysql.pm
```diff
@@ -171,7 +171,8 @@ has _mysql => (
 
         # Special-case --password, which requires = before the value. O_o
         if (my $pw = $self->password) {
-            push @ret, "--password=$pw";
```
theory: The password in `$pw` can be different from the one in the file, so I think this should probably be:

``` diff
 
         # Special-case --password, which requires = before the value. O_o
         if (my $pw = $self->password) {
-            push @ret, "--password=$pw";
+            my $cfgpw = $self->_mycnf->{password} || '';
+            push @ret, "--password=$pw" if $pw ne $cfgpwd;
         }
 
         # Options to keep things quiet.
```

kielstr: I have added this change ;)

PR #508, opened 2020-04-21
show.pm missing Target

clad:
Hi,

Hit dependency issue with show command when using -t option.

Can't locate object method "new" via package "App::Sqitch::Target" at /usr/local/share/perl/5.28.1/App/Sqitch/Command/show.pm line 47.

PR adds missing line.

Thanks,
Clinton


clad:
will redo for develop branch

PR #512, opened 2020-05-02
Check command

theory:
Updates and replaces #477.


theory:
Looks like we missed some spots for the script hash; added them in 2bf83083.

PR #549, opened 2020-09-30
Update snowflake.pm

john-auld:
Update of the Snowflake URL pattern to include privatelink addresses.

https://docs.snowflake.com/en/user-guide/admin-security-privatelink.html

### DIFF
--- a/lib/App/Sqitch/Engine/snowflake.pm
+++ b/lib/App/Sqitch/Engine/snowflake.pm
@@ -134,6 +134,7 @@ has account => (
         my $self = shift;
         if (my $host = $self->uri->host) {
             # <account_name>.<region_id>.<cloud_platform>.snowflakecomputing.com
+            # <account_name>.<region_id>.privatelink.snowflakecomputing.com (<cloud_platform> is not included, even if the non-privatelink address contains that var)
             $host =~ s/[.]snowflakecomputing[.]com$//;
             return $host;
         }


theory:
> <cloud_platform> is not included, even if the non-privatelink address contains that var

What does this mean?

john-auld:
> > <cloud_platform> is not included, even if the non-privatelink address contains that var
> 
> What does this mean?

A special endpoint can be created inside an AWS and Azure VPC to avoid the need for the clients in the VPC to route connections for Snowflake out over the Internet. For a new Snowflake account the URL for connecting over the Internet, without using PrivateLink, the pattern of the URL would be

**account.region.account_platform.snowflakecomputing.com**.

When connecting to the same account inside of a VPC using PrivateLink, the URL would be

**account.region.privatelink.snowflakecomputing.com**.

john-auld:
P.S. Your code update solves a problem we have with privateLink in the v1.1.0 tag.

theory:
But then isn't `privatelink` just one of the options for the `<cloud_platform>` variable?

And is the "code update" you reference f17e5a4?

john-auld:
`privatelink` represents the manner of the connection over a secure endpoint and it can be used with AWS or Azure so it has a separate meaning. However, the parsing rule added in commit f17e5a4 (line 137) work with privatelink.

A more succinct way of expressing the new URL format for Snowflake is given below.

`# <account_name>.<region_id>.<cloud_platform||privatelink>.snowflakecomputing.com`

PR #555, opened 2020-12-13
Fix sqlite3 inplace changes tutorial

TheGhostHuCodes:
When making an inplace change to the userflips view I think we should modify the
`revert/userflips.sql` file, not the `revert/userflips@v1.0.0-dev2.sql`
that is generated for us by sqitch.


theory:
Quite right, nice spot, thank you!

Issue #579, opened 2021-06-22
Race condition when executing multiple copies of sqitch on the same DB


neilmayhew:
Our app is deployed in multiple identical containers to form a load-balanced cluster. Each container needs to ensure that DB migrations are run before the app starts up, so we do this in the entrypoint script. It's not possible to predict which container will run first, so they all do it. The sqitch documentation suggests that this is a safe thing to do because of the locking done by `begin_work`. However, it doesn't work properly.

I'm able to reproduce the problem in development by running `sqitch deploy` from two different terminals. If the sqitch in the first terminal has already started on the plan, the second one says

```
Change "App::Sqitch::Plan::Change=HASH(0x4f543f0)" has already been deployed
```

and exits, even though not all of the changes have been deployed yet. (And in production the app then starts up and fails immediately because it doesn't match the DB schema.)

Once I improved the message to show a meaningful ID I was able to see what's happening. The second sqitch figures out which changes remain to be deployed, and does sanity checking on them in `check_deploy_dependencies`. However, by this time the first sqitch has deployed another change and the second sqitch aborts because it's trying to deploy an already-deployed change.

In production we don't hit this all the time because we're typically deploying only a single new change to an existing DB, and the chance of overlap is low. However, in our testing environments we start with an empty DB and apply all the migrations since the beginning of time in a big block. This triggers the problem almost every time.

I was surprised that the second sqitch doesn't just block until the first one is finished deploying all the changes, but I see that the locking happens around each individual change and not the entire series. As a result, querying and deploying can end up interleaved. This seems like a design error to me, and I feel that sqitch should take the locks before it does anything at all, and not release them until it's ready to exit.

What do you think? I'm interested in submitting a PR for this, but I don't want to spend time on it if it's unlikely to be accepted.

theory:
Yeah, I think what happens it that it calls `begin_work` table before each change, rather than all changes. I have to admit it never occurred to me that deployment would be automated across a number of services restarting. In my mind, the deployment of the database is separate from the services that use it, but obviously sometimes they are not compatible across versions of the services that use the database. So the process it typically:

1. Shut down services
2. Update the database (once in one place)
3. Upgrade the services
4. Start the upgraded services

[...]

PR #589, opened 2021-09-12
Add destination locking to pgsql and mysql

theory:
This PR takes on the issue raised in #579 by adding support for advisory locking to the engine, with implementations for MySQL and Postgres. It relies on engines to implement two methods:

* `try_lock`, which will try to create an exclusive lock and immediately return on success or failure
* `wait_lock`, which will try to create an exclusive lock and wait indefinitely to acquire it

Both methods should create exactly the same lock, statically defined by the engines, so that no other instance of Sqitch can be making changes to the database at the same time. The new `lock_destination` method is called at the start of both `deploy` and `revert`. It calls `try_lock`, and if it succeeds, returns, but if it doesn't, it emits a message about its inability to get the lock, then calls `wait_lock` to wait for it. The new `lock_timeout` attribute controls the wait timeout (60s by default), and `lock_destination` handles the timeout by raising an error, which will cause Sqitch to exit.

New tests in `t/engine.t` cover the behavior of `lock_destination`, while new testing and configuration tests engine-specific implementations in `t/lib/DBIEngineTest`.

I'd appreciate a close reading of this PR, as there are a few things we should consider. In particular:

* Do we really want this enabled by default? Or should we add configuration/options to turn it on? Or perhaps configuration/options to turn it off?
* Is using Sys::SigAction sufficient to handle the timeout, including on non-Unix hosts (Windows)? Or is it likely to run into issues/conflicts with the database? Might some database locking carry on even when the timeout kicks in? (I suspect it does, since I had to add code to repeatedly clear locks to the tests, though in practice it should not be an issue, since Sqitch will exit, not hang around holding onto locks unnecessarily. But MySQL in particular is weird).

I expect to add the necessary configuration/options in a followup PR. Thank you!

Closing #579.

### DIFF
--- a/Changes
+++ b/Changes
@@ -36,6 +36,14 @@ Revision history for Perl extension App::Sqitch
      - Fixed the error message when attempting to deploy a change that has
        already been deployed to display the name of the change, rather than a
        memory address. Thanks to Neil Mayhew for the report (#579).
+     - Added destination locking, currently implemented for PostgresQL and
+       MySQL. On starting a deploy or revert, Sqitch attempts to "lock the
+       destination" using advisory locks, to ensure that only one instance of
+       Sqitch makes changes to the database at any one time. This complements
+       the existing locking, which applies as each change is deployed or
+       reverted, as that pattern led to failures when multiple instances of
+       Sqitch were working at once. Thanks to Neil Mayhew for the report
+       (#579).
 
 1.1.0  2020-05-17T16:20:07Z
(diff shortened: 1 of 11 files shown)


pdl:
(I appreciate this isn't what you asked for but flagging it so it doesn't get lost)
Why the magic number 75474063 for postgres? Is that a number which happens to be the oid of a sqitch schema/relation in your db?

theory:
I used random.org to generate a number. That's it, I'm boring, sorry.

PR #596, opened 2021-11-20
Add several additional dzil plugins

autarch:
* CopyFilesFromBuild - this copies the generated cpanfile back into the repo.
* Git::Check - checks that the repo is in a good state for release, and doesn't have any uncommitted files (with a few exceptions).
* Git::Commit - this will commit the generated cpanfile back to the distro after the release.
* Git::Tag - tags the release in git.
* Git::Push - pushes the commit and tag back to origin.


theory:
> * CopyFilesFromBuild - this copies the generated cpanfile back into the repo.
> * Git::Check - checks that the repo is in a good state for release, and doesn't have any uncommitted files (with a few exceptions).
> * Git::Commit - this will commit the generated cpanfile back to the distro after the release.
> * Git::Tag - tags the release in git.
> * Git::Push - pushes the commit and tag back to origin.

Awesome, thank you. However, one of the GitHub workflows I added was a release workflow. It's triggered when a tag starting with `v` is pushed. Am I right in thinking that the last two items might conflict with that pattern?

autarch:
We can remove the `Git::Tag` part entirely. The real issue is `Git::Push`. That's needed so that the generated `cpanfile` is pushed back to the repo. For `Git::Push`, it looks like all's that needed is setting the git user and email to something - https://github.community/t/push-built-files-to-repo-using-actions/18186

theory:
Okay, let's do that.

autarch:
> Okay, let's do that.

Alright, I force pushed a new version with that change.

PR #604, opened 2022-01-17
Skip OS tests on tag.

theory:
All the workflows aside from release.yml have a branch spec now, so none will run when pushing a tag.


theory:
Relevant [SO answer](https://stackoverflow.com/a/70743841/79202) (thx @dvarrazzo).

ssoriche (approved):
👍

PR #621, opened 2022-03-27
Fix a Firebird test and upgrade issues

### LINKED ISSUE #597: t/firebird.t fails
If DBD::Firebird is installed then t/firebird.t seems to fail (seen on centos8, fedora and debian11 systems, various perl versions):
```
    #   Failed test 'Database should not yet be initialized'
    #   at t/lib/DBIEngineTest.pm line 103.
    # Looks like you failed 1 test of 2.

#   Failed test 'live database'
#   at t/lib/DBIEngineTest.pm line 1780.
Sqitch database /var/tmp/cpansmoker-1023/2021112112/__metasqitch already initialized

Trace begun at /home/cpansand/.cpan/build/2021112112/App-Sqitch-v1.2.0-2/blib/lib/App/Sqitch/Engine/firebird.pm line 206
App::Sqitch::Engine::firebird::initialize('App::Sqitch::Engine::firebird=HASH(0x53b1ee0)') called at t/lib/DBIEngineTest.pm line 108
DBIEngineTest::__ANON__ at /opt/perl-5.28.2t/lib/site_perl/5.28.2/Test/Builder.pm line 374
eval {...} at /opt/perl-5.28.2t/lib/site_perl/5.28.2/Test/Builder.pm line 374
Test::Builder::subtest('Test::Builder=HASH(0x20aa6c0)', 'live database', 'CODE(0x535ae58)') called at /opt/perl-5.28.2t/lib/site_perl/5.28.2/Test/More.pm line 809
Test::More::subtest('live database', 'CODE(0x535ae58)') called at t/lib/DBIEngineTest.pm line 1780
DBIEngineTest::run('DBIEngineTest', 'class', 'App::Sqitch::Engine::firebird', 'target_params', 'ARRAY(0x4825a40)', 'alt_target_params', 'ARRAY(0x482d6b8)', 'skip_unless', 'CODE(0x482d0e8)', 'engine_err_regex', 'Regexp=REGEXP(0x5373018)', 'init_error', 'Sqitch database /var/tmp/cpansmoker-1023/2021112112/__sqitchtest already initialized', 'add_second_format', 'dateadd(1 second to %s)', 'test_dbh', 'CODE(0x4517c78)') called at t/firebird.t line 381
Unsuccessful execution caused by a system error that precludes successful execution of subsequent statements
-I/O error during "open" operation for file "/var/tmp/cpansmoker-1023/2021112112/__sqitchtest"
-Error while trying to open file
-No such file or directory at t/firebird.t line 323.
END failed--call queue aborted.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 82.
t/firebird.t ........ 
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 1/82 subtests 
```

theory:
Commit bdc2b0caf conflated the target database with the registry database. So give the registry a more distinctive name, restore the separate name for the database we create for testing (when none exists already), and skip cleanup for any database files that don't exist (because no live tests were run).

Also, disconnect from a local Firebird connection on upgrade. With a direct connection (with no host name), there can be only one connection do the database at a time. This is a problem for upgrades, where we already have a DBI
connection and then use `isql` to perform the upgrade. So disconnect DBI from the database before performing an upgrade, and add a Moo method to clear the `dbh` attribute, which will allow it to reconnect the next time it's called.

Figured out by using the Docker image specified in #597 and installing the `firebird3.0-utils` package.

Fixes #597.

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,6 +2,11 @@ Revision history for Perl extension App::Sqitch
 
 1.2.2
 
+     - Fixed an issue when testing Firebird on a host with Firebird installed
+       but no `isql`, and when using a local Firebird (e.g., the Engine12
+       provider), which allows only one connection at a time. Thanks to Slaven
+       Rezić for the the reproducible configuration (#597).
+
 1.2.1  2021-12-05T19:59:45Z
      - Updated all the live engine tests, aside from Oracle, to test with
        unique registry names, so as to avoid conflicts when multiple
--- a/lib/App/Sqitch/Engine/firebird.pm
+++ b/lib/App/Sqitch/Engine/firebird.pm
@@ -66,6 +66,7 @@ has dbh => (
     is      => 'rw',
     isa     => DBH,
     lazy    => 1,
+    clearer => '_clear_dbh',
     default => sub {
         my $self = shift;
         my $uri  = $self->registry_uri;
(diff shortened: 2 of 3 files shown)


theory:
Rebased and pushed in 5f483db7.

PR #625, opened 2022-04-19
Fix duplicated headers on sqitch.plan tutorial.

alysson-azevedo:
In the tutorial, after merging the branches, the header in the example is duplicated. It should appear only once.


theory:
Thank you!

PR #635, opened 2022-05-28
Add support for YugabyteDB

theory:
Works great with the Postgres engine except for locks. So detect that it's Yugabyte and don't create the locks if it is. Change the handling of the `lock_sql` option in the database tests to take a code reference instead of a hash reference, and only return the locking instructions in `t/pg.t` if the engine is not Yugabyte. Mention the Yugabyte support here and there.

Alway sort the current state test tags for Yugabyte, as previously done only for Oracle and MySQL. The Postgres aggregate docs say that the ordering of `array_agg()` is undefined, though it has been reliable in all testing. Not so for Yugabyte. So we force the sort for Yugabyte, but also added #636 to propose a better path forward for aggregate sorting

Add GitHub Actions testing for Yugabyte 2.6 and higher, and link it in the README. While at it, clean up the README, properly listing all supported engines and linking to them.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,11 +1,13 @@
 Revision history for Perl extension App::Sqitch
 
 1.2.2
-
      - Fixed an issue when testing Firebird on a host with Firebird installed
        but no `isql`, and when using a local Firebird (e.g., the Engine12
        provider), which allows only one connection at a time. Thanks to Slaven
        Rezić for the the reproducible configuration (#597).
+     - Tweaked the Postgres engine to support Yugabyte. The only unsupported
+       features are explicit locks, so users need to manually ensure that only
+       one instance of Sqitch is updating the cluster at a time.
 
 1.2.1  2021-12-05T19:59:45Z
      - Updated all the live engine tests, aside from Oracle, to test with
--- a/lib/App/Sqitch/Engine/pg.pm
+++ b/lib/App/Sqitch/Engine/pg.pm
@@ -11,6 +11,7 @@ use Locale::TextDomain qw(App-Sqitch);
 use App::Sqitch::Plan::Change;
 use List::Util qw(first);
 use App::Sqitch::Types qw(DBH ArrayRef);
+use Type::Utils qw(enum);
 use namespace::autoclean;
 
 extends 'App::Sqitch::Engine';
@@ -101,6 +102,13 @@ sub name   { 'PostgreSQL' }
 sub driver { 'DBD::Pg 2.0' }
 sub default_client { 'psql' }
 
+has _provider => (
+    is      => 'rw',
+    isa     => enum([qw( postgres yugabyte )]),
+    default => 'postgres',
+    lazy    => 1,
+);
+
 has dbh => (
     is      => 'rw',
     isa     => DBH,
@@ -135,6 +143,12 @@ has dbh => (
                         # https://www.nntp.perl.org/group/perl.dbi.dev/2013/11/msg7622.html
                         $dbh->set_err(undef, undef) if $dbh->err;
                     };
+                    # Determine the provider. Yugabyte says this is the right way to do it.
+                    # https://yugabyte-db.slack.com/archives/CG0KQF0GG/p1653762283847589
+                    my $v = $dbh->selectcol_arrayref(
+                        q{SELECT split_part(version(), ' ', 2)}
+                    )->[0] // '';
+                    $self->_provider('yugabyte') if $v =~ /-YB-/;
                     return;
                 },
             },
@@ -167,6 +181,10 @@ sub _char2ts { $_[1]->as_string(format => 'iso') }
 
 sub _listagg_format {
     q{ARRAY(SELECT * FROM UNNEST( array_agg(%s) ) a WHERE a IS NOT NULL)}
+    
+    # Consider dropping 8.4 and passing the sort by column to ensure that list
+    # is always in a well-defined order:
+    # q{ARRAY(SELECT * FROM UNNEST( array_agg(%s ORDER BY %s) ) a WHERE a IS NOT NULL)}
 }
 
 sub _regex_op { '~' }
@@ -259,7 +277,10 @@ sub begin_work {
 
     # Start transaction and lock changes to allow only one change at a time.
     $dbh->begin_work;
-    $dbh->do('LOCK TABLE changes IN EXCLUSIVE MODE');
+    $dbh->do('LOCK TABLE changes IN EXCLUSIVE MODE')
+        unless $self->_provider eq 'yugabyte';
+        # Yugabyte does not yet support EXCLUSIVE MODE.
+        # https://docs.yugabyte.com/preview/api/ysql/the-sql-language/statements/txn_lock/#lockmode-1
     return $self;
 }
 
@@ -274,6 +295,14 @@ sub try_lock {
 # until timeout.
 sub wait_lock {
     my $self = shift;
+
+    # Yugabyte and Cockroach do not support advisory locks.
+    # https://github.com/yugabyte/yugabyte-db/issues/3642
+    # https://github.com/cockroachdb/cockroach/issues/13546
+    # Use pessimistic locking when it becomes available.
+    # https://github.com/yugabyte/yugabyte-db/issues/5680
+    return 1 if $self->_provider ne 'postgres';
+
     # Asynchronously request a lock with an indefinite wait.
     my $dbh = $self->dbh;
     $dbh->do(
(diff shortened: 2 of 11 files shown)


Changes
```diff
@@ -1,11 +1,13 @@
 Revision history for Perl extension App::Sqitch
 
 1.2.2
-
      - Fixed an issue when testing Firebird on a host with Firebird installed
        but no `isql`, and when using a local Firebird (e.g., the Engine12
        provider), which allows only one connection at a time. Thanks to Slaven
        Rezić for the the reproducible configuration (#597).
+     - Tweaked the Postgres engine to support Yugabyte. The only features
+       not supported are locking, so users need to manually sure tha only one
```
autarch, line 9: "ensure that"

lib/App/Sqitch/Engine/pg.pm
```diff
@@ -135,6 +143,12 @@ has dbh => (
                         # https://www.nntp.perl.org/group/perl.dbi.dev/2013/11/msg7622.html
                         $dbh->set_err(undef, undef) if $dbh->err;
                     };
+                    # Determine the provider. Yugabyte says this is the righ tway to do it.
```
autarch, line 146: "right way"

autarch (changes requested):
This looks good to me. Just a few typos in comments.

theory:
Thanks @autarch, fixed those in b901fe8c.

stefanoric:
Hey @theory how is this supposed to be used? I am testing sqitch with yugabyte and still getting the advisory lock error.

```
$> sqitch deploy db:yugabyte://yugabyte@127.0.0.11:5433/yugabyte
ERROR:  advisory locks are not yet implemented
HINT:  See https://github.com/yugabyte/yugabyte-db/issues/3642. React with thumbs up to raise its priority

$> ✗ sqitch --version
sqitch (App::Sqitch) v1.4.1

$> brew info yugabytedb
==> yugabyte/tap/yugabytedb: stable 2024.1.3.0

**Comment by stefanoric** [comment]
I get this:
```
 PostgreSQL 11.2-YB-2.21.0.0-b0 on x86_64-apple-darwin21.6.0, compiled by Apple clang version 14.0.0 (clang-1400.0.29.202), 64-bit
```

**Comment by theory** [comment]
Hrm. What's the output of

```sql
SELECT split_part(version(), ' ', 2);
```

It _should_ be `11.2-YB-2.21.0.0-b0`, in which case `$v =~ /-YB-/` should match.

stefanoric:
Yes, this is what I get:

```
yugabyte@yugabyte=# select split_part(version(), ' ',2);
     split_part
─────────────────────
 11.2-YB-2.21.0.0-b0
(1 row)
```

FranckPachot:
@theory  At that time, I think the error was a warning, so maybe the provider check never worked.
@stefanoric As a workaround, it is possible to ignore the error with `set yb_silence_advisory_locks_not_supported_error to true`

stefanoric:
@FranckPachot thank you for chiming in. I am not entirely sure where that statement is meant to be issued. I don't have a session open on the database when the sqitch scripts run. I have tried setting this as my first DDL statement, but it's not effective (likely sqitch is trying to create the lock _before_ starting to run the schema creation steps).

theory:
Oh, has calling `pg_try_advisory_lock()` changed from a warning to an error? If so, this bit probably needs updating:

https://github.com/sqitchers/sqitch/blob/a7dfdb3b16979bf9cdd414af10eb1a7236a36066/lib/App/Sqitch/Engine/pg.pm#L294-L298

It already skips calling `pg_advisory_lock()` a few lines below:

https://github.com/sqitchers/sqitch/blob/a7dfdb3b16979bf9cdd414af10eb1a7236a36066/lib/App/Sqitch/Engine/pg.pm#L305-L309

theory:
@stefanoric If it's a connection parameter, try adding it to the target URI:

```sh
sqitch deploy "db:pg://yugabyte@127.0.0.11:5433/yugabyte? yb_silence_advisory_locks_not_supported_error=true"
```

theory:
Or try this patch:

```diff
--- a/lib/App/Sqitch/Engine/pg.pm
+++ b/lib/App/Sqitch/Engine/pg.pm
@@ -292,7 +292,13 @@ sub begin_work {
 
 # Override to try to acquire a lock on a constant number without waiting.
 sub try_lock {
-    shift->dbh->selectcol_arrayref(
+    my $self = shift;
+    # Yugabyte does not support advisory locks.
+    # https://github.com/yugabyte/yugabyte-db/issues/3642
+    # Use pessimistic locking when it becomes available.
+    # https://github.com/yugabyte/yugabyte-db/issues/5680
+    return 1 if $self->_provider ne 'postgres';
+    $self->dbh->selectcol_arrayref(
         'SELECT pg_try_advisory_lock(75474063)'
     )->[0]
 }
```

stefanoric:
It does not seem to be possible to append that command to the connection string, I get a `no matches found`.

stefanoric:
The patch did not work either, I am still getting the advisory lock error (I believe I have applied it correctly because if I put a syntax error there it breaks, so it's going through that code).

theory:
Weird. What does 

```pl
print $self->_provider, "\n";
```

output there?

stefanoric:
This is what I get: 
```
✗ sqitch deploy db:pg://yugabyte@127.0.0.11:5433/yugabyte
postgres
ERROR:  advisory locks are not yet implemented
HINT:  See https://github.com/yugabyte/yugabyte-db/issues/3642. React with thumbs up to raise its priority
```

theory:
Okay! That shows that it's not detecting Yugabyte. We can fix that! I've opened #841 for this issue; let's move discussion there. Would you post the version you're running?

FranckPachot:
> It does not seem to be possible to append that command to the connection string, I get a `no matches found`.

other solution:
```
export PGOPTIONS="-c yb_silence_advisory_locks_not_supported_error=on"
```
before calling `sqitch`

FranckPachot:
@theory the detection is ok. With my limited perl memories, I got
```
The provider is: yugabyte
```
with
```
                    # Determine the provider. Yugabyte says this is the right way to do it.
                    # https://yugabyte-db.slack.com/archives/CG0KQF0GG/p1653762283847589
                    my $v = $dbh->selectcol_arrayref(
                        q{SELECT split_part(version(), ' ', 2)}
                    )->[0] // '';
                    $self->_provider('yugabyte') if $v =~ /-YB-/;
                    
                    my $provider = $self->_provider();  
                    print STDERR "The provider is: $provider\n";
```

theory:
Thank you, @FranckPachot. Just need to wonder why @stefanoric gets `postgres`.

stefanoric:
```
sqitch deploy db:yugabyte://yugabyte@127.0.0.11:5433/yugabyte
The provider at line 298 is: postgres
The provider at line 150 is: yugabyte
ERROR:  advisory locks are not yet implemented
HINT:  See https://github.com/yugabyte/yugabyte-db/issues/3642. React with thumbs up to raise its priority
```

stefanoric:
> > It does not seem to be possible to append that command to the connection string, I get a `no matches found`.
> 
> other solution:
> 
> ```
> export PGOPTIONS="-c yb_silence_advisory_locks_not_supported_error=on"
> ```
> 
> before calling `sqitch`

@FranckPachot  This gets me a bit farther, but eventually, I get this:

```
✗ sqitch deploy db:pg://yugabyte@127.0.0.11:5433/yugabyte
The provider at line 298 is: postgres
The provider at line 150 is: yugabyte
Adding registry tables to db:pg://yugabyte@127.0.0.11:5433/yugabyte
"ysqlsh " failed to start: "No such file or directory"
```

I am not sure where that reference to `ysqlsh` is coming from.

FranckPachot:
`ysqlsh` is the rename of `psql` in the YugabyteDB binaries, but I guess `sqitch` is using `psql`
No idea where it comes from. Especially with a space in it `"ysqlsh "`

FranckPachot:
Maybe you have set it in `.conf` instead of `psql`?
```
[engine "pg"]
    client    = /usr/local/pgsql/bin/psql
```

stefanoric:
Ah... that was it! Some remnant of my trying things :) 

I finally got it to work. 

It's expected that it's failing, as I wanted to test reverts (since my understanding is that Yugabyte does not allow to roll back a DDL definition like Postgres does. Individual steps in Sqitch may be a good approximation.
```
sqitch deploy db:pg://yugabyte@127.0.0.11:5433/shortener
The provider at line 298 is: postgres
The provider at line 150 is: yugabyte
Adding registry tables to db:pg://yugabyte@127.0.0.11:5433/shortener
Deploying changes to db:pg://yugabyte@127.0.0.11:5433/shortener
  + initial-schema ....... ok
  + create-token-tables .. ok
  + test-error-step ...... psql:deploy/test-error-step.sql:5: ERROR:  table "tokens_unexisting" does not exist
not ok
"psql" unexpectedly returned exit value 3

Reverting all changes
  - create-token-tables .. ok
  - initial-schema ....... ok
Deploy failed
```

theory:
> ```
> sqitch deploy db:yugabyte://yugabyte@127.0.0.11:5433/yugabyte
> The provider at line 298 is: postgres
> The provider at line 150 is: yugabyte
> ERROR:  advisory locks are not yet implemented
> HINT:  See https://github.com/yugabyte/yugabyte-db/issues/3642. React with thumbs up to raise its priority
> ```

That's just wild!

PR #658, opened 2022-07-22
Add fractional seconds on mariadb for versions 5.03.05 or higher

rbrigot:


theory:
So sub-second precision in was added in MariaDB at 50305? Seems like a perfectly reasonable change, thank you!

theory:
> So sub-second precision in was added in MariaDB at 50305?

Duh, that's just what the name of this PR says 🤦🏻

theory (approved):

theory:
Subsumed by #660.

PR #674, opened 2022-08-20
Explicitly require DBD::Mem in tests

### LINKED ISSUE #673: install_driver(Mem) failed: Can't locate DBD/Mem.pm in @INC
* http://www.cpantesters.org/cpan/report/c28c5b36-1aca-11ed-b5dd-e77c2cbfd570
* http://www.cpantesters.org/cpan/report/04f57dfb-6bfa-1014-8121-85be8c60feb7
* http://www.cpantesters.org/cpan/report/b7796e8c-1aca-11ed-b5dd-e77c2cbfd570

Pretty odd since DBD::Mom is distributed with DBI, but whatever. Will add it as an explicit requirement.

theory:
Fixes #673.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -86,6 +86,9 @@ Time::Local = 0
 [Prereqs / DevelopSuggests]
 DBD::Oracle = 1.23
 
+[Prereqs / TestRequires]
+DBD::Mem = 0
+
 ;; Recommend author dependencies (dzil) in develop/recommends.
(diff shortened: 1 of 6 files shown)


theory:
@eserte have you had a chance to try this?

theory:
> @eserte have you had a chance to try this?

Oops, sorry, mean the patch in #673, which is what would be generated on building the dist from this patch.

PR #676, opened 2022-08-20
Do not pass the port if it is not in the URI

### LINKED ISSUE #675: `PGPORT` environment variable doesn't seem to be respected in v1.3.0
Hi there guys, wanted to alert you to an issue that I started facing starting after the v1.3.0 release on Friday. I use sqitch via the docker image & the recommended script as documented in the README and docker hub. I am not sure if this an issue with the v1.3.0 release in general or just the docker setup. I'm also not sure if it's all `PG*` environment variables or just `PGPORT`. Basically, the behavior I'm seeing is that even though `PGPORT` is in the environment, when trying to run migrations I get an error like this:
```
Adding registry tables to pg-localdev
psql: error: could not connect to server: Connection refused
        Is the server running on host "localhost" (::1) and accepting
        TCP/IP connections on port 5432?
could not connect to server: Connection refused
        Is the server running on host "localhost" (127.0.0.1) and accepting
        TCP/IP connections on port 5432?
"psql" unexpectedly returned exit value 2
```
Steps to reproduce:
1. Have a postgres instance running on port 5433
2. Download the `sqitch` docker running script from the README of this repo
3. Try to run some migrations with `PGPORT=5433` in the environment

This is what my sqitch config file looks like:
```
[core]
  engine = pg
  top_dir = db/
[engine "pg"]
  target = pg-localdev
[target "pg-localdev"]
  uri = db:postgresql://postgres:password@localhost/database
```

I was able to solve this issue for myself by just pinning the tag in the `sqitch` runner script to `v1.2.1.1` (side note: `v1.2.1` is listed as a supported tag on docker hub but doesn't seem to work: https://hub.docker.com/r/sqitch/sqitch/ ; I had to come look at the tags in the repo to find that you use a versioning scheme w/ 4 components instead of 3). So I'm not blocked here or anything, just thought you guys might want a heads up as this caused some github actions to start failing for me; I am using their "port mapping" functionality (https://docs.github.com/en/actions/using-containerized-services/about-service-containers) and was specifying the port with `PGPORT`, and my tests suddenly started failing. I'm guessing the same might be true for some others as well.

theory:
Otherwise PGPORT is overridden! Fixes #675.

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,6 +2,10 @@ Revision history for Perl extension App::Sqitch
 
 1.3.1
 
+     - Fixed a bug introduced in v1.3.0 where the Postgres engine would
+       always pass the port to `psql`, thus ignoring the `PGPORT` environment
+       variable. Thanks to Cam Feenstra for the spot (#675)!
+
 1.3.0  2022-08-12T22:09:13Z
      - Fixed an issue when testing Firebird on a host with Firebird installed
        but no `isql`, and when using a local Firebird (e.g., the Engine12
--- a/lib/App/Sqitch/Engine/pg.pm
+++ b/lib/App/Sqitch/Engine/pg.pm
@@ -57,11 +57,13 @@ has _psql => (
 
         my %query_params = $uri->query_params;
         my @conninfo;
+        # Use _port instead of port so it's empty if no port is in the URI.
+        # https://github.com/sqitchers/sqitch/issues/675
         for my $spec (
             [ user   => $self->username ],
             [ dbname => $uri->dbname    ],
             [ host   => $uri->host      ],
-            [ port   => $uri->port      ],
+            [ port   => $uri->_port     ],
             map { [ $_ => $query_params{$_} ] }
                 sort keys %query_params,
         ) {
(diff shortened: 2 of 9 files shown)


Changes
```diff
@@ -2,6 +2,10 @@ Revision history for Perl extension App::Sqitch
 
 1.3.1
 
+     - Fixed a bug introduced in v1.3.0 where the Postgres engine would ignore
```
autarch, line 5: "ignore always". I think you just mean "always.

lib/App/Sqitch/Engine/pg.pm
```diff
@@ -61,7 +61,7 @@ has _psql => (
             [ user   => $self->username ],
             [ dbname => $uri->dbname    ],
             [ host   => $uri->host      ],
-            [ port   => $uri->port      ],
+            [ port   => $uri->_port      ],
```
autarch, line 64: It might be worth adding a comment that `_port` is in fact the documented way to get a port-sans-default from a `URI` (gross API).

autarch (changes requested):
Looks good except for some small issues.

lib/App/Sqitch/Engine/pg.pm
theory: Good idea, avoid the issue in the future (hopefully). Done in 8f72ee6a.

autarch (approved):

PR #677, opened 2022-08-21
Use database name in MySQL lock

### LINKED ISSUE #670: MySQL deploy blocked by get_lock on another schema
Since get_lock from

> `lib/App/Sqitch/Engine/mysql.pm:310:`        q{SELECT get_lock('sqitch working', 0)}`

Instantiates serverwide exclusive lock, there is no possibility to run simultaneous deploy on different schemas. 

Also, the error message

>  Blocked by another instance of Sqitch working on {TARGET}; waiting 60 seconds...

 appears to be misleading, and displays the target name of current deploy, not the one that holds the lock
 
 steps to reproduce:
 create a target DBs, users and grants:
```create schema t1;
create schema t1_sqitch;

create schema t2;
create schema t2_sqitch;

create user t1;
grant all privileges on t1.* to t1;
grant all privileges on t1_sqitch.* to t1;

create user t2;
grant all privileges on t2.* to t2;
grant all privileges on t2_sqitch.* to t2;
```
 
 create two sqtich projects, init sqitch,  add targets and add a target per project so sqitch.config looks as follows:
```[core]
        engine = mysql
        plan_file = sqitch.plan
        # top_dir = .
# [engine "mysql"]
        # target = db:mysql:
        # registry = sqitch
        # client = mysql
[target "t1"]
        uri = db:mysql://t1@127.0.0.1:3306/t1
        registry = t1_sqitch
```

```[core]
        engine = mysql
        plan_file = sqitch.plan
        # top_dir = .
# [engine "mysql"]
        # target = db:mysql:
        # registry = sqitch
        # client = mysql
[target "t2"]
        uri = db:mysql://t2@127.0.0.1:3306/t2
        registry = t2_sqitch
```

add sample migration with sqitch add for each project 
```
BEGIN;
select sleep(90);
COMMIT;
``` 

 run sqitch in two terminal windows in parallel 
 project1 `sqitch deploy -t t1`
 project2 `sqitch deploy -t t2`
 
 while the first run works, the second one returns an error:

```
Blocked by another instance of Sqitch working on t2; waiting 60 seconds...
Timed out waiting 60 seconds for another instance of Sqitch to finish work on t2
```

theory:
So that the scope of the lock is limited only to that database. This is probably the most intuitive way to do it, in that it's probably what the user expects. Preventing parallel deploys to different databases is more confusing. Resolves #670.

### DIFF
--- a/lib/App/Sqitch/Engine/mysql.pm
+++ b/lib/App/Sqitch/Engine/mysql.pm
@@ -313,24 +313,35 @@ sub begin_work {
     return $self;
 }
 
-# Override to try to acquire a lock on the string "sqitch working" without
-# waiting.
+# We include the database name in the lock name because that's probably the most
+# stringent lock the user expects. Locking the whole server with a static string
+# prevents parallel deploys to other databases. Yes, locking just the target
+# allows parallel deploys to conflict with one another if they make changes to
+# other databases, but is not a great practice and likely an anti-pattern. So
+# stick with the least surprising behavior.
+# https://github.com/sqitchers/sqitch/issues/670
+sub _lock_name {
+    'sqitch working on ' . shift->uri->dbname
+}
+
+# Override to try to acquire a lock on the string "sqitch working on $dbname"
+# without waiting.
 sub try_lock {
     my $self = shift;
     # Can't create a lock in the registry if it doesn't exist.
     $self->initialize unless $self->initialized;
     $self->dbh->selectcol_arrayref(
-        q{SELECT get_lock('sqitch working', 0)}
+        q{SELECT get_lock(?, ?)}, undef, $self->_lock_name, 0,
     )->[0]
 }
 
-# Override to try to acquire a lock on the string "sqitch working", waiting
-# for the lock until timeout.
+# Override to try to acquire a lock on the string "sqitch working on $dbname",
+# waiting for the lock until timeout.
 sub wait_lock {
     my $self = shift;
     $self->dbh->selectcol_arrayref(
-        q{SELECT get_lock('sqitch working', ?)},
-        undef, $self->lock_timeout
+        q{SELECT get_lock(?, ?)}, undef,
+        $self->_lock_name, $self->lock_timeout,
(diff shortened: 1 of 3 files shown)


lib/App/Sqitch/Engine/mysql.pm
```diff
@@ -313,24 +313,35 @@ sub begin_work {
     return $self;
 }
 
-# Override to try to acquire a lock on the string "sqitch working" without
-# waiting.
+# We include the database name in the lock name because that's probably the most
+# stringent lock the user expects. Locking the whole server with a static string
+# prevents parallel deploys to other databases. Yes, locking just the target
+# allows parallel deploys to conflict with one another if they make changes to
+# other databases, but is not a great practice and likely an exceptional
```
autarch, line 320: "this is not"

lib/App/Sqitch/Engine/mysql.pm
```diff
@@ -313,24 +313,35 @@ sub begin_work {
...
+# pattern. So stick with the least surprising behavior.
```
autarch, line 321: "anti-pattern", I assume.

autarch (changes requested):
Looks good except for some typos in comments.

lib/App/Sqitch/Engine/mysql.pm
theory: It follows "locking just the target", so "locking just the target is not" is grammatically correct.

lib/App/Sqitch/Engine/mysql.pm
theory: Yeah that's better, changed in 914b44cd

lib/App/Sqitch/Engine/mysql.pm
autarch: Right. But I think in that case you don't want a comment before the "but".

theory:
Really appreciate you taking the time to look over Sqitch PRs, @autarch. You make them better.

PR #688, opened 2022-11-20
Quote Snowflake identifiers

### LINKED ISSUE #685: Snowflake engine does not support fulle range of characters in role/warehouse identifiers
If you use special characters like dot "." in either a role or a warehouse name, SQITCH will fail executing USE ROLE <role_name>, or USE WAREHOUSE <warehouse_name>.

As stated on [Snowflake docs](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html), identifiers can include a wide range of characters in object identifiers if they are quoted.

So far, the Sqitch snowflake engine supports unquoted identifiers only.
https://github.com/sqitchers/sqitch/blob/8f72ee6a34029c18228ed4b30ad2b239ee8760e3/lib/App/Sqitch/Engine/snowflake.pm#L214-L217

The solution could be :
- quoting the role or warehouse name in the code above systematically, with a small regression for users who did not set their SQITCH role/warehouse names with the right case (quoted identifiers are case sensistive, unquoted ones case insensitive)
- detect if the role/warehouse names have to be quoted or not, and quote them or not accordingly.

theory:
Resolves #685

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for Perl extension App::Sqitch
 
 1.3.2
+     - Fixed Snowflake warehouse and role setup to properly quote identifiers.
+       Thanks to @marc-marketparts for the report (#685)!
 
 1.3.1  2022-10-01T18:49:30Z
      - Fixed a bug introduced in v1.3.0 where the Postgres engine would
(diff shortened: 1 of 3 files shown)


autarch (approved):
LGTM!

theory:
Thanks Deve, appreciate it!

PR #689, opened 2022-11-20
Create rework directory if it does not exist

### LINKED ISSUE #686: Running `rework` with a subdirectory that has not been manually created fails
While using Sqitch version 1.3.1, I have a plan that has added `tables/users` in a previous tag. There have been no reworks in this project. When I run `sqitch rework -m "Add stuff to users" tables/users` then I get an error and no files are generated. The error is this:

```
Cannot copy sql/revert/tables/users.sql to sql/_reworked/revert/tables/users@1.0.0.sql: No such file or directory
```

Expected behaviour: Sqitch should create directories and files that it needs to make the rework command succeed.

Workaround: I can create the directories manually: `mkdir -p sql/_reworked/deploy/tables sql/_reworked/revert/tables sql/_reworked/verify/tables`

theory:
Add tests to make sure it works as expected. Resolves #686.

### DIFF
--- a/Changes
+++ b/Changes
@@ -3,6 +3,8 @@ Revision history for Perl extension App::Sqitch
 1.3.2
      - Fixed Snowflake warehouse and role setup to properly quote identifiers.
        Thanks to @marc-marketparts for the report (#685)!
+     - Fixed a bug reworking a change when a rework directory is configured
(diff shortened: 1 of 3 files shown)


autarch (dismissed):
LGTM!

theory:
Thanks Dave, always appreciate another set of eyes.

PR #702, opened 2023-01-11
Print changes to be deployed or reverted.

### LINKED ISSUE #701: Wishlist: Print the list of changes that would be deployed or reverted
Currently sqitch will print something like "Deploying changes through <change>" but it would be good to see a list of the actual changes that would be deployed. This is especially powerful when combined with the revert.prompt_accept functionality.

@theory If there is buy-in for this idea I am happy to write a patch.

vectro:
Closes #701

### DIFF
--- a/lib/App/Sqitch/Engine.pm
+++ b/lib/App/Sqitch/Engine.pm
@@ -249,6 +249,11 @@ sub deploy {
         )
     );
 
+    $sqitch->debug(__ "Will deploy the following changes:");
+    foreach my $will_deploy_position ($plan->position .. $to_index) {
+        $sqitch->debug($plan->change_at($will_deploy_position)->format_name_with_tags);
+    }
+
     # Check that all dependencies will be satisfied.
     $self->check_deploy_dependencies($plan, $to_index);
 
@@ -267,6 +272,15 @@ sub deploy {
     );
 
     $self->$meth( $plan, $to_index );
+}
+
+# Do a thing similar to Sqitch::Plan::Change::format_name_with_tags,
+# but for an output from $self->deployed_changes or
+# $self->deployed_changes_since.
+sub _format_deployed_change_name_with_tags($) {
+    my ( $self, $change ) = @_;
+
+    return join ' ', $change->{name}, map { '@' . $_ } @{$change->{tags}};
 }
 
 sub revert {
@@ -299,6 +313,8 @@ sub revert {
             );
         };
 
+        # NB this is an array of unblessed references, not of
+        # Sqitch::Plan::Change references.
         @changes = $self->deployed_changes_since(
             $self->_load_changes($change)
         ) or do {
@@ -308,6 +324,8 @@ sub revert {
             ));
             return $self;
         };
+        my @change_descriptions =
+            map { $self->_format_deployed_change_name_with_tags($_) } @changes;
 
         if ($self->no_prompt) {
             $sqitch->info(__x(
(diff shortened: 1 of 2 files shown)


lib/App/Sqitch/Engine.pm
```diff
@@ -249,6 +249,11 @@ sub deploy {
         )
     );
 
+    $sqitch->info(__ "Will deploy the following changes:");
```
theory, line 252: I think this should be tied to a flag. Or maybe just emit this information if `$sqitch->verbosity > 1`) (1 is the default).

lib/App/Sqitch/Engine.pm
```diff
@@ -269,6 +274,15 @@ sub deploy {
     $self->$meth( $plan, $to_index );
 }
 
+# Do a thing similar to Sqitch::Plan::Change::format_name_with_tags,
```
theory, line 277: If you can't use that actual method, I suggest adding a new one there, not here.

t/engine.t
```diff
@@ -1159,8 +1171,8 @@ is_deeply $engine->seen, [
 ], 'Should have deployed to change 2';
 is_deeply +MockOutput->get_info, [
     [__x 'Deploying changes to {destination}', destination =>  $engine->destination ],
-    # [__ 'ok'],
```
theory, line 1162: Him, why would these go away?

theory (changes requested):
Looking good, but a couple changes I think would make it just right. Thanks!

t/engine.t
vectro: I have no objection to leaving them but do you mind if I ask why they are there?

lib/App/Sqitch/Engine.pm
vectro: We do not have a Sqitch::Plan::Change here, so I don't think this code can go in that module?

lib/App/Sqitch/Engine.pm
vectro: How is this?

t/engine.t
theory: Oh, I mis-read them as output (some output starts with `#`), but since they're commented out, yeah, fine to remove them.

lib/App/Sqitch/Engine.pm
theory: Ah, the `debug` method, very good!

lib/App/Sqitch/Engine.pm
theory: Oh 🤦🏻 got it.

theory: I do wonder if we should avoid the overhead when verbosity is less than two. Though it probably isn't that expensive, relative to the overhead of the database. Maybe I'm overthinking it.

theory (approved):

vectro:
Regarding performance, my assumption would be that the overhead in printing would be orders of magnitude smaller than the actual time to deploy the change, even for trivial changes.

vectro:
Also I'm not sure what the next step is on this PR, do you want to run workflows? Is it ready to merge?

theory:
Yeah, I'm overthinking it on performance :-) I've approved the workflows; is there a test that demonstrates that those lines do not output when verbosity is < 2? If so I think it's good to go.

vectro:
> is there a test that demonstrates that those lines do not output when verbosity is < 2?

If you are referring to the `deploy` change, there is a test, in the sense that we never print this message in tests. All the tests only check verbosity = 1.

If you are referring to the `revert` change, we print these at verbosity = 1.

vectro:
Regarding these CI failures, I'm not sure what to make of them. They don't seem related to this change? Are the tests a bit flaky?

theory:
Yeah I probably need to re-enable Snowflake and figure out what Exasol's problem is.

vectro:
What is the next step on this PR? Is it ready to merge?

theory:
I'll take a look at getting it merged soon, hopefully this weekend, but maybe not for a week or two.

PR #703, opened 2023-01-18
Allow false separators in formatters

web-vertalo:
Seemingly valid separators like empty string and 0 that evaluate to false result in use of the default separator. Test for defined separator instead of true.

Tests pass before and after this update. Apologies for not including new tests.

### DIFF
--- a/lib/App/Sqitch/ItemFormatter.pm
+++ b/lib/App/Sqitch/ItemFormatter.pm
@@ -128,12 +128,12 @@ has formatter => (
 
                 t => sub {
                     @{ $_[0]->{tags} }
-                        ? ' ' . join $_[1] || ', ' => @{ $_[0]->{tags} }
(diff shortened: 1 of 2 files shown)


theory (changes requested):
Thank you for the PR! This looks pretty straightforward, though I'm dubious on the use of an empty string as a separator.  Would you mind adding a test demonstrating the use of empty or `0` separators? If empty space looks hinky, maybe use `length` instead of `defined` to allow `0`.

web-vertalo:
Tests added. My take on the empty separator is that it may make sense for tags, and may make no sense elsewhere. If caller specifies it, then reverting to default separators violates the principle of least surprise and adds an opinion to sqitch. So I'd leave it to the caller.

theory (approved):
Fair enough I take your point. Thank you for the contribution!

PR #704, opened 2023-01-19
Refactor revert options

vectro:
Refactor the way we manage the no_prompt and prompt_accept configuration options.

Specifically change these from being a mutable part of engine state to values passed into the `revert` method directly.

This PR is preparation for addressing some of the ideas in #659.

This is a refactor only, there is no behavior change. However some tests that call directly into particular APIs have been adjusted to use the new APIs.

### DIFF
--- a/lib/App/Sqitch/Engine.pm
+++ b/lib/App/Sqitch/Engine.pm
@@ -67,6 +67,28 @@ has start_at => (
     isa => Str
 );
 
+has no_prompt => (
+    is      => 'rw',
+    isa     => Bool,
+    trigger => sub {
+        # Deprecation notice added Feb 2023
+        warnings::warnif("deprecated",
+                         __("Engine::no_prompt is deprecated and will be removed in a future release\n".
+                            "Use direct arguments to revert() instead"));
+    }
+);
+
+has prompt_accept => (
+    is      => 'rw',
+    isa     => Bool,
+    trigger => sub {
+        # Deprecation notice added Feb 2023
+        warnings::warnif("deprecated",
+                         __("Engine::prompt_accept is deprecated and will be removed in a future release\n".
+                            "Use direct arguments to revert() instead"));
+    }
+);
+
 has log_only => (
     is      => 'rw',
     isa     => Bool,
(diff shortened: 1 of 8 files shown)


lib/App/Sqitch/Engine.pm
```diff
@@ -269,8 +257,17 @@ sub deploy {
...
-sub revert {
-    my ( $self, $to ) = @_;
+sub revert($$$$) {
+    # $to = revert up to (but not including) this change. May be undefined.
+    # $prompt = If true, we ask for confirmation; if false, we don't.
+    # $prompt_default = Default if the user just hits enter at the prompt.
+    my ( $self, $to, $prompt, $prompt_default ) = @_;
+
+    if (defined $prompt) {
+        hurl revert => 'Missing mandatory parameter $prompt_default' unless defined $prompt_default;
```
theory, line 267: These should be internationalized by using, IIRC,

``` perl
 hurl revert => __x(
    'Missing mandatory parameter {param}',
    param => '$prompt_default',
) unless defined $prompt_default;
```

lib/App/Sqitch/Engine.pm
```diff
@@ -1572,10 +1572,6 @@ The name of the registry schema or database.
 
 The point in the plan from which to start deploying changes.
 
-=head3 C<no_prompt>
-
-Boolean indicating whether or not to prompt for reverts. False by default.
-
```
theory, line 1578: Ugh, did we never document `prompt_accept`?

theory (changes requested):
Thanks for the PR! Can you explain a bit about why these changes are necessary in order to implement the feature described in #659?

Otherwise, a couple things:

1. Removing those attributes might break things for any other tools that use this module. There don't appear to be any [on MetaCPAN[(https://metacpan.org/module/App::Sqitch::Engine/requires), but it's the sort of thing people write tools against internally. Still, better if we can retain the compatibility (hence my question above)
2. Please use localization `__`, `__x`, etc.) for all messages. 

Thanks!

lib/App/Sqitch/Engine.pm
vectro: It seems that way?

vectro:
Hmm okay. I didn't understand that this was intended to be a stable interface.

Regarding the motivation for this change: As part of implementing #659, I wanted a prompt option for deploy that would let users see what would be deployed before actually deploying it. To prepare for that, I thought it would be better if the `no_prompt` and `prompt_accept` parameters were passed by value as method arguments instead of by mutable state, because otherwise it wouldn't be clear whether the mutable state was intended to apply to the revert or deploy commands (or both). Or to put it another way, one could set no_prompt at one point and have it unexpectedly apply to a later call to `deploy`.

If after reading this you think this change is a mistake, then we can close this PR and I can proceed without this refactor.

theory:
Hrm, you make good points about the clarity of purpose. Maybe it'd make sense to leave the attributes for compatibility, and then support optional arguments to the methods to override those attributes. Would that work?

lib/App/Sqitch/Engine.pm
vectro: Agreed, though $prompt_default is not a variable here, we want to print a literal `$`.

vectro:
How do you feel about this approach? The new arguments are optional but not using them generates a deprecation warning.

theory:
That works. I wasn't aware of `warnings::warnif`!

theory (approved):
Looks good, thanks!

PR #705, opened 2023-01-27
Add MySQL 5.7 support

vladsf:
Add fixes to support MySQL 5.7

### DIFF
--- a/lib/sqitchtutorial-mysql.pod
+++ b/lib/sqitchtutorial-mysql.pod
@@ -85,7 +85,7 @@ as appropriate.
 By default, Sqitch will read F<sqitch.conf> in the current directory for
 settings. But it will also read F<~/.sqitch/sqitch.conf> for user-specific
 settings. Since MySQL's
-L<C<mysql> client|https://dev.mysql.com/doc/refman/5.6/en/mysql.html> is not
+L<C<mysql client|https://dev.mysql.com/doc/refman/5.6/en/mysql.html> is not
 in the path on my system, let's go ahead an tell it where to find the client
 on our computer (don't bother if you're using the
 L<Docker image|https://hub.docker.com/r/sqitch/sqitch/> because it uses the
--- a/lib/sqitchtutorial-mysql.pod
+++ b/lib/sqitchtutorial-mysql.pod
@@ -85,7 +85,7 @@ as appropriate.
 By default, Sqitch will read F<sqitch.conf> in the current directory for
 settings. But it will also read F<~/.sqitch/sqitch.conf> for user-specific
 settings. Since MySQL's
-L<C<mysql client|https://dev.mysql.com/doc/refman/5.7/en/mysql.html> is not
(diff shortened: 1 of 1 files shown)


theory:
Thanks. I wonder if we need to include instructions for older versions of MySQL, too? Or is 8 old enough now that we can just standardize on it? If the latter, would you mind adding a note near the top that the examples assume MySQL 8? Thanks!

vladsf:
MySQL 8 is old enough. Released 5 years ago. It does not make much sense to
promote MySQL 5.x in the tutorial.

vladsf:
I found another problem: ENCRYPT() function was deprecated in MySQL 5.7 and removed in MySQL 8.0.  SHA2() was recommended to use (https://dev.mysql.com/doc/refman/5.7/en/encryption-functions.html#function_sha2) instead.

vladsf:
`ENCRYPT()` call might be replaced with:
`SHA2(CONCAT(HEX(RANDOM_BYTES(2)), 'password'), 256)`
the the whole code will be 8.0 compatible otherwise it requires 5.[6,7]

theory:
> `ENCRYPT()` call might be replaced with:
> `SHA2(CONCAT(HEX(RANDOM_BYTES(2)), 'password'), 256)`
> then the whole code will be 8.0 compatible otherwise it requires 5.[6,7]

Sounds good, if you want to add that I'll get it merged. Thank you!

vladsf:
Removing ENCRYPT() breaks the consistency of tutorial. Some filenames has _encrypt_ in the names, etc. I suggest to stay compatible with MySQL 5.7. This version is available on the official MySQL Docker page.

vladsf:
@theory any objections?

vladsf:
@theory any updates?

theory:
Nope, been completely heads-down on a project at work, plus a mini-project to move the PGXN bot from Twitter to Mastodon. I'll get to it, but it might be a couple weeks.

lib/sqitchtutorial-mysql.pod
```diff
@@ -85,7 +85,7 @@ as appropriate.
 By default, Sqitch will read F<sqitch.conf> in the current directory for
 settings. But it will also read F<~/.sqitch/sqitch.conf> for user-specific
 settings. Since MySQL's
-L<C<mysql> client|https://dev.mysql.com/doc/refman/5.6/en/mysql.html> is not
+L<C<mysql client|https://dev.mysql.com/doc/refman/5.7/en/mysql.html> is not
```
theory, line 88: The closing `>` after `mysql` is queried to complete `C<mysql>`.

theory (dismissed):
Looks good, just one Pod bit that needs fixing.

lib/sqitchtutorial-mysql.pod
vladsf: OK, fixed. Enclosed the whole "mysql client".

PR #716, opened 2023-02-20
Update checkout & cache actions

theory:
And use the new environment files for setting output for a release tarabll. Done in response to warnings in workflow runs pointing to these anouncements:

  https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
  https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


autarch (approved):
LGTM!

PR #719, opened 2023-02-28
Add a new strict-mode configuration option

vectro:
This new configuration option disables some functionality that could be perceived as creating risks to production environments.

To begin with, that includes:

- The checkout and rebase commands
- Running revert without a change specified ("Revert All Changes" function).

### DIFF
--- a/lib/App/Sqitch/Command/revert.pm
+++ b/lib/App/Sqitch/Command/revert.pm
@@ -43,6 +43,16 @@ has prompt_accept => (
     isa => Bool
 );
 
+has strict => (
+    is       => 'ro',
+    lazy     => 1,
+    default  => sub {
+        my $self = shift;
+        return ($self->sqitch->config->get( key => 'revert.strict' )
+                // 0);
+    }
+);
+
 has log_only => (
     is       => 'ro',
     isa      => Bool,
@@ -134,10 +144,29 @@ sub execute {
     my $change = $self->modified
         ? $engine->planned_deployed_common_ancestor_id
         : $self->to_change // shift @{ $changes };
-    $self->warn(__x(
-        'Too many changes specified; reverting to "{change}"',
-        change => $change,
-    )) if @{ $changes };
+
+    if (!defined $change && $self->strict) {
+        hurl {
+            ident   => 'revert:strict',
+            message => __ 'Must specify a target revision in strict mode',
+            exitval => 1,
+        };
+    }
+
+    if (@{ $changes }) {
+        if ($self->strict) {
+            hurl {
+                ident   => 'revert:strict:excess',
+                message => __ 'Too many changes specified',
+                exitval => 1,
+            };
+        }
+
+        $self->warn(__x(
+            'Too many changes specified; reverting to "{change}"',
+            change => $change,
+        ));
+    }
 
     # Now get to work.
     $engine->log_only( $self->log_only );
@@ -202,6 +231,11 @@ the revert.
 
 Boolean value to indicate whether or not the default value for the prompt,
 should the user hit C<return>, is to accept the prompt or deny it.
+
+=head3 C<strict>
+
+Boolean indicating whether or not specification of a target change is
+mandatory.
 
 =head3 C<target>
 
--- a/lib/App/Sqitch/Role/RevertDeployCommand.pm
+++ b/lib/App/Sqitch/Role/RevertDeployCommand.pm
@@ -8,6 +8,8 @@ use Moo::Role;
 use App::Sqitch::Types qw(Str Int Bool HashRef);
 use Type::Utils qw(enum);
 use namespace::autoclean;
+use Locale::TextDomain qw(App-Sqitch);
+use App::Sqitch::X qw(hurl);
 
 requires 'sqitch';
 requires 'command';
@@ -51,6 +53,22 @@ has no_prompt => (
 has prompt_accept => (
     is  => 'ro',
     isa => Bool
+);
+
+has strict => (
+    is       => 'ro',
+    lazy     => 1,
+    default  => sub {
+        my $self = shift;
+        my $cmd = $self->command;
+        return ($self->sqitch->config->get(
+                    key => "$cmd.strict",
+                    as  => 'bool',
+                ) // $self->sqitch->config->get(
+                    key => 'revert.strict',
+                    as  => 'bool',
+                ) // 0);
+    }
 );
 
 has mode => (
@@ -174,6 +192,22 @@ around configure => sub {
     return $params;
 };
 
+sub BUILD {
+    my ($self, $args) = @_;
+
+    if ($self->strict) {
+        hurl {
+            ident   => 'lax_command',
+            exitval => 1,
+            message => __x(
+                '"{command}" cannot be used when strict mode is enabled.\n'.
+                'Consider using revert and deploy commands directly.',
+                command => $self->command,
+                ),
+        };
+    }
+};
+
 1;
 
 __END__
@@ -227,6 +261,11 @@ the revert.
 
 Boolean value to indicate whether or not the default value for the prompt,
 should the user hit C<return>, is to accept the prompt or deny it.
+
+=head3 C<strict>
+
+Boolean value to indicate whether or not strict mode is enabled; if
+so, use of these commands is prohibited.
 
 =head3 C<target>
 
--- a/t/rebase.t
+++ b/t/rebase.t
@@ -714,4 +714,41 @@ for my $spec (
         "Should rethrow $spec->[0] exception";
 }
 
+# Should die if running in strict mode.
+ok $config = TestConfig->new(
+    'revert.strict'    => 1
+), 'Create strict config';
+ok $sqitch = App::Sqitch->new(config => $config),
+    'Load a sqitch sqitch object';
+throws_ok {
+    $CLASS->new(
+        sqitch           => $sqitch,
+        ); }
+    'App::Sqitch::X',
+    'Cannot initialize command in strict mode.';
+
+ok $config = TestConfig->new(
+    'rebase.strict'    => 1
+), 'Create strict config';
+ok $sqitch = App::Sqitch->new(config => $config),
+    'Load a sqitch sqitch object';
+throws_ok {
+    $CLASS->new(
+        sqitch           => $sqitch,
+        ); }
+    'App::Sqitch::X',
+    'Cannot initialize command in strict mode.';
+
+# rebase.strict overrides revert.strict
+ok $config = TestConfig->new(
+    'revert.strict'    => 1,
+    'rebase.strict'    => 0
+), 'Create strict config';
+ok $sqitch = App::Sqitch->new(config => $config),
+    'Load a sqitch sqitch object';
+ok $CLASS->new(
+    sqitch           => $sqitch,
+    ),
+   'Okay to initialize because rebase is not in strict mode';
+
 done_testing;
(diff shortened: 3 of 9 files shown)


lib/App/Sqitch.pm
```diff
@@ -176,6 +176,17 @@ has pager => (
     },
 );
 
+has strict => (
```
theory, line 179: Since the option applies only to deploys and reverts, IIUC, I think it'd be better placed in `RevertDeployCommand.pm`. Would require updating `around options` and `around configure` in that file.

t/rebase.t
```diff
@@ -15,6 +15,7 @@ use Test::Warn;
 use lib 't/lib';
 use MockOutput;
 use TestConfig;
+use Carp::Always;
```
theory, line 18: I don't think you need this if you're using `hurl` and as long as the tests test for warnings.

lib/sqitch-config.pod
```diff
@@ -484,6 +484,25 @@ by C<core.top_dir>.
 The file name extension on deploy, revert, and verify SQL scripts. Defaults to
 C<sql>.
 
+=item C<core.strict>
+
+A boolean value indicating whether or not to disallow certain
+potentially risky operations. Defaults to false.
+
+When enabled:
```
theory, line 492: Can you add tests for these conditions? I can't tell at a glance how some commands are allowed.

lib/App/Sqitch/Role/RevertDeployCommand.pm
```diff
@@ -174,6 +176,22 @@ around configure => sub {
     return $params;
 };
 
+sub BUILD {
+    my ($self, $args) = @_;
+
+    if ($self->sqitch->strict) {
```
theory, line 182: Does this not prevent any deploy or revert at all?

lib/sqitch-config.pod
```diff
@@ -484,6 +484,25 @@ by C<core.top_dir>.
 The file name extension on deploy, revert, and verify SQL scripts. Defaults to
 C<sql>.
 
+=item C<core.strict>
```
theory, line 487: Appreciate the docs here, but I think the option and config need to be documented for every command they apply to.  Should include, at least, these files, I think:

```
lib/sqitch-deploy.pod
lib/sqitch-revert-usage.pod
lib/sqitch-rebase.pod
lib/sqitch-revert.pod
lib/sqitch-deploy-usage.pod
lib/sqitch-checkout-usage.pod
lib/sqitch-checkout.pod
lib/sqitch-rebase-usage.pod
lib/App/Sqitch/Command/revert.pm
lib/App/Sqitch/Command/deploy.pm
```

The main pod files for each command also include a "Configuration Variables" that should have the strict mode documented -- anywhere the string `no_prompt` appears.

theory (changes requested):

lib/App/Sqitch.pm
vectro: It also applies to just `revert`, which doesn't inherit from `RevertDeployCommand`, or at least that's how I understand it.

If this were in `RevertDeployCommand`, could we make the change to `Engine.pm` in this PR?

t/rebase.t
vectro: Yes, agreed, this snuck in somehow.

lib/sqitch-config.pod
vectro: Could you rephrase this request? I think the PR already adds tests for the behavior documented here.

lib/App/Sqitch/Role/RevertDeployCommand.pm
vectro: This will completely block the use of `checkout` and `rebase`, which inherit from this module; but not `deploy` or `revert` since they do not inherit from this module.

Or have I misunderstood your question?

lib/sqitch-config.pod
vectro: I'm happy to extend the documentation of course, but in some of the cases that you listed I'm not sure what I would say.
* `sqitch-deploy.pod`, `sqitch-deploy-usage.pod`, and `deploy.pm`: I don't think this change affects the deploy command at all?
* `sqitch-deploy-usage.pod`, `sqitch-checkout-usage.pod`, `sqitch-revert-usage.pod`, and `sqitch-rebase-usage.pod`: These files don't seem to currently describe configuration options at all. Are you proposing to add new sections?
* `deploy.pm` and `revert.pm`: It looks to me like these are only documenting the attributes, not the configuration, is that wrong? I don't think this PR changes any of the attributes for these two modules.

>The main pod files for each command also include a "Configuration Variables" that should have the strict mode documented -- anywhere the string no_prompt appears.

I think this is covered by your suggestions above, but let me know if you disagree.

I will push a change to `sqitch-rebase.pod`, `sqitch-revert.pod`, `sqitch-checkout.pod` in the meantime.

lib/App/Sqitch.pm
theory: Right, but it's not an option for all or even most commands. A similar pattern exists for `no_prompt` and `prompt_accept`: They're defined both in RevertDeployCommand and separately in revert.pm. I think we should stick to that pattern and not make it a global option.

lib/sqitch-config.pod
theory: * Right, not deploy, sorry. 
* I see options listed in the `-usage.pod` files; new ones  like `--strict` that apply to each command should be added (this is related to my other request not to make it a global option I guess).
* Yeah, if you move the option to `revert.pm` it would then need to be documented.

lib/sqitch-config.pod
theory: I see it now, sorry.

lib/App/Sqitch/Role/RevertDeployCommand.pm
theory: Ah, I see, missed the inheritance. It has been a minute since I spent time with this code. :-)

lib/sqitch-config.pod
vectro: Oh, I don't think we are adding new options here, only new configuration variables.

Unless there is something which I don't understand that automatically creates options from configuration settings?

lib/App/Sqitch/Command/revert.pm
```diff
@@ -134,10 +144,28 @@ sub execute {
...
+        };
+    }
+    elsif (@{ $changes }) {
+        if ($self->strict) {
+            hurl {
+                ident   => 'revert:strict:excess',
+                message => __ 'Too many changes specified',
+                exitval => 1,
+            };
+        } else {
```
theory, line 162: Don't need the `else` here; `hurl` raises an exception.

lib/sqitch-config.pod
theory: Oh, right. In that case it should probably follow the cascading configuration pattern of

1. `--strict`
2. `$cmd.strict`
3. `target.$name.strict`
4. `engine.$name.strict`
5. `core.strict`

Most of that resolution is done in App::Sqitch::Target, using the `_fetch` method:

https://github.com/sqitchers/sqitch/blob/aace1e837b5e07fa0d8cfb924505b221f58079c4/lib/App/Sqitch/Target.pm#L76-L84

I think it'd be fine to do so in a subsequent PR.

theory (approved):

lib/App/Sqitch/Command/revert.pm
vectro: How's this?

theory (commented):
Okay, did some thinking through of this proposal again, and here's where I'm at:

* Like `no_prompt`, `strict` should be configured as a revert variable only, `revert.strict`. And just as `no_prompt` can be set by `-y`, so too can `strict` be set by `--strict`, but also by `--no-strict`. There is no need for `core.strict` because it applies to no other actions other than revert.
* Why not allow the `rebase` command to succeed when strict is enabled as long as `to_change` is set?
* The checkout command always has an implicit target: the most recent common change between the current and target branches. Why disable it?
* The `--strict` option must be documented for every command where it's relevant. Based on the `-y` option, that's:
    *   lib/sqitch-checkout-usage.pod
    *   lib/sqitch-checkout.pod
    *   lib/sqitch-rebase-usage.pod
    *   lib/sqitch-rebase.pod
    *   lib/sqitch-revert-usage.pod
    *   lib/sqitch-revert.pod
* Likewise, the `revert.strict` config variable should be documented anywhere `revert.no_prompt` is documented, which appears to be:
    *   lib/sqitch-rebase.pod
    *   lib/sqitch-revert.pod
    *   lib/sqitch-checkout.pod
    *   lib/App/Sqitch/Role/RevertDeployCommand.pm
    *   lib/App/Sqitch/Command/revert.pm

Depending on where we come down on the affect on `rebase` and `checkout`, we may want to also add support for `rebase.strict` and `checkout.strict`, again just like `no_prompt`, but that may be excessive.

What do you think?

vectro:
Hi David,

Do you mind saying more about the use case that you imagine for the --strict and --no-strict command line options?

In my mind, the purpose of the strict mode is to prevent the accidental use of potentially dangerous features, either by users who are not familiar with Sqitch, or by users who might be familiar with these commands but targeting the wrong environment. In other words, it's to protect users from accidentally doing the wrong thing.

From that perspective I don't understand the situation where someone would want to enable or disable this on the command line. If you want the strict behavior, I would think you especially want it in situations where people didn't think to ask for it.

Ian

theory:
Ah yes, I was working off the assumption that almost every configuration in Sqitch also has an option; overlooked that your PR does not.

In that case, though, I think it makes sense to follow the `prompt_accept` configuration, which does not have an equivalent option, and is set via `revert.prompt_accept' (no `core.prompt_accept`), and overridden by `checkout.prompt_accept` or `rebase.prompt_accept`.

It might be useful to extend these two options to targets, too, as I mentioned earlier, but not necessary here.

vectro:
> It might be useful to extend these two options to targets

I agree, but maybe we can do that in a subsequent MR?

theory:
> I agree, but maybe we can do that in a subsequent MR?

Yes, absolutely.

vectro:
>Why not allow the rebase command to succeed when strict is enabled as long as to_change is set?

I guess my belief is that this command is good for rapid development but that when it comes to production deployments (where reverting can cause data loss) it would be better to be more explicit about what changes should be reverted. The whole idea of reverting and then re-deploying seems to me like an exceptional workflow in a production environment — even though it could happen from time to time, it's better to be explicit about it.

(Assuming you meant `onto-change`, let me know if I misunderstood the suggestion)

vectro:
>The checkout command always has an implicit target: the most recent common change between the current and target branches. Why disable it?

It's less about the possibility that all changes would be reverted and more about the possibility than anything could be reverted without the operator having the specific intention to revert those specific changes. In other words, "implied target change" is probably usually better than "revert everything", but still risky.

vectro:
@theory What do you think about this latest attempt?

theory:
> (Assuming you meant `onto-change`, let me know if I misunderstood the suggestion)

Yes, that's what I meant, was thinking about it from the POV of consistency. This is part of why I've been thinking it'd be useful to support the option at the target level, as you would want strict enabled for a prod target but not a dev target. That said, we can leave it as `revert.strict` and the disabling behavior, and recommend that it be used at the `--user` or `--system` config level. You'd enable it on whatever box was responsible for production reverts.

theory:
> @theory What do you think about this latest attempt?

Looks good, I think it's almost there. I'll pull it down and give it a through review this weekend. Thanks!

theory:
Have a look at #735, @vectro. If it looks good to you I'll merge it (which will also merge this PR). Thanks!

theory:
Closing in favor of #735, which includes your commit.

PR #735, opened 2023-04-03
Refactor strict mode 

theory:
Build on #719. Remove the attribute from RevertDeployCommand and handle it in the configure method. Remove redundant logic from revert.pm. Tweak the docs and remove the incorrect statement that `-y` overrides strict mode. Note the new feature in Changes and credit @vectro. Also fix a few typos in test descriptions.

### DIFF
--- a/lib/App/Sqitch/Command/revert.pm
+++ b/lib/App/Sqitch/Command/revert.pm
@@ -43,6 +43,16 @@ has prompt_accept => (
     isa => Bool
 );
 
+has strict => (
+    is       => 'ro',
+    lazy     => 1,
+    default  => sub {
+        my $self = shift;
+        return ($self->sqitch->config->get( key => 'revert.strict' )
+                // 0);
+    }
+);
+
 has log_only => (
     is       => 'ro',
     isa      => Bool,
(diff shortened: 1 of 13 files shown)


lib/App/Sqitch/Command/revert.pm
```diff
@@ -145,28 +144,17 @@ sub execute {
...
-
-        $self->warn(__x(
-            'Too many changes specified; reverting to "{change}"',
-            change => $change,
-        ));
-    }
+    $self->warn(__x(
+        'Too many changes specified; reverting to "{change}"',
+        change => $change,
+    )) if @{ $changes };
```
vectro: I think strict mode should make it a fatal error to specify too many changes, do you disagree?

My rationale is that in strict mode, when reverting the user should be specific about what changes should be reverted, and ignoring a user-specified change goes against that principle.

vectro (changes requested):
Looks good to me, other than my one other comment.

lib/App/Sqitch/Command/revert.pm
theory: Oh yeah, I was thinking it was fatal, but it's a warning. Will need to add a test for that.

theory: Fixed in e439eb79.

theory:
Rebased onto develop in cb939e95.

PR #745, opened 2023-05-03
File name formatting option

web-vertalo:
This PR adds a format option %F to sqitch plan that prints the path of the file used to deploy each migration in the plan. The initial use case is to search DB source code in reverse order of deployment. This makes it easy to explore the history of a function or table.

The code prints the full path, including the directory prefix and the .sql suffix, but it's trivial to modify those with an external tool, so we kept the update simple.

### DIFF
--- a/lib/App/Sqitch/Command/plan.pm
+++ b/lib/App/Sqitch/Command/plan.pm
@@ -236,6 +236,7 @@ sub execute {
             change_id     => $change->id,
             change        => $change->name,
             note          => $change->note,
+            deploy_file   => $change->deploy_file,
             tags          => [ map { $_->format_name } $change->tags ],
             requires      => [ map { $_->as_string } $change->requires ],
             conflicts     => [ map { $_->as_string } $change->conflicts ],
--- a/lib/App/Sqitch/ItemFormatter.pm
+++ b/lib/App/Sqitch/ItemFormatter.pm
@@ -105,6 +105,7 @@ has formatter => (
                 },
                 n => sub { $_[0]->{change} },
                 o => sub { $_[0]->{project} },
+                F => sub { $_[0]->{deploy_file} },
 
                 c => sub {
                     return "$_[0]->{committer_name} <$_[0]->{committer_email}>"
@@ -439,6 +440,8 @@ The placeholders are:
 =item * C<%n>: Event change name
 
 =item * C<%o>: Event change project name
+
+=item * C<%F>: Deploy file name
 
 =item * C<%($len)h>: abbreviated change of length C<$len>
 
--- a/lib/sqitch-plan.pod
+++ b/lib/sqitch-plan.pod
@@ -294,6 +294,8 @@ The placeholders are:
 
 =item * C<%o>: Event change project name
 
+=item * C<%F>: Deploy file name
+
 =item * C<%($len)h>: abbreviated change of length C<$len>
 
 =item * C<%e>: Event type (deploy, revert, fail)
--- a/Changes
+++ b/Changes
@@ -30,6 +30,8 @@ Revision history for Perl extension App::Sqitch
        (libwww-perl/URI#13). Thanks to @bobfang for the report (#744)!
      - Fixed test failures due to a change in the generation of DBI DSN by
        URI::Oracle introduced by libwww-perl/URI-db#23.
+     - Added a format option `%F` to sqitch plan that prints the path of the
+       file used to deploy each migration in the plan.
 
 1.3.1  2022-10-01T18:49:30Z
      - Fixed a bug introduced in v1.3.0 where the Postgres engine would
--- a/t/item_formatter.t
+++ b/t/item_formatter.t
@@ -3,7 +3,7 @@
 use strict;
 use warnings;
 use utf8;
-use Test::More tests => 170;
+use Test::More tests => 172;
 #use Test::More 'no_plan';
 use App::Sqitch;
 use Locale::TextDomain qw(App-Sqitch);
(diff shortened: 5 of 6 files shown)


lib/App/Sqitch/Plan/Change.pm
```diff
@@ -221,6 +221,11 @@ sub verify_file {
     $self->verify_dir->file( $self->path_segments );
 }
 
+sub file {
+    my $self = shift;
+    return $self->is_deploy ? $self->deploy_file : $self->revert_file;
+}
+
```
theory, line 228: I don't think this is necessary; the `deploy_file` is always deployed, even if the action is a revert. The `revert_file` is used to revert the change — even if the change is a revert!

That said, the specification of revert changes, while supported by the plan file syntax from near the beginning, has never been implemented, and I'm don't recall ever working out how it should work, really. So I think you're best off using `deploy_file` exclusively.

lib/App/Sqitch/Command/plan.pm
```diff
@@ -236,6 +236,7 @@ sub execute {
             change_id     => $change->id,
             change        => $change->name,
             note          => $change->note,
+            file          => $change->file,
```
theory, line 239: I guess it's not a direct change to add it to the `log` command, too. Bit of a pity, since the `log` and `plan` commands otherwise support exactly the same formats.

lib/App/Sqitch/ItemFormatter.pm
```diff
@@ -440,6 +441,8 @@ The placeholders are:
 
 =item * C<%o>: Event change project name
 
+=item * C<%F>: Event file name
```
theory, line 444: Suggest "deploy file name".

lib/sqitch-plan.pod
```diff
@@ -294,6 +294,8 @@ The placeholders are:
 
 =item * C<%o>: Event change project name
 
+=item * C<%F>: Event file name
```
theory, line 297: Suggest "deploy file name".

theory (changes requested):
This looks useful, but doesn't `%{deploy_file}a` do the trick as well?

web-vertalo:
> This looks useful, but doesn't `%{deploy_file}a` do the trick as well?

That results in `deploy_file is not a valid change attribute`, but `%{file}a` works. Knowing that would have saved a bunch of effort. How can one determine the list of valid attributes?

lib/App/Sqitch/ItemFormatter.pm
web-vertalo: Updated.

lib/App/Sqitch/Plan/Change.pm
web-vertalo: Updated.

web-vertalo:
> > This looks useful, but doesn't `%{deploy_file}a` do the trick as well?
> 
> That results in `deploy_file is not a valid change attribute`, but `%{file}a` works. Knowing that would have saved a bunch of effort. How can one determine the list of valid attributes?

The list is in Command/plan.pm line 233.

web-vertalo:
> > This looks useful, but doesn't `%{deploy_file}a` do the trick as well?
> 
> That results in `deploy_file is not a valid change attribute`, but `%{file}a` works. Knowing that would have saved a bunch of effort. How can one determine the list of valid attributes?

The list is in Command/plan.pm line 233.

lib/App/Sqitch/Command/plan.pm
ghoust-xxx: I looked in it, but decided not to, because on this moment it lacks some of `plan` format options anyway. Like `%r` or `%t` for example (though they are in documentation). They are in `ItemFormatter.pm` (it is the same class for both `plan` and `log`), but values for them do not passed in `log.pm`.

web-vertalo:
Of course @ghoust-xxx is correct. I like having the documented %F shorthand and propose to keep the PR as is.

ghoust-xxx:
@theory so, can you accept this PR, or we have to do some changes?

theory:
> @theory so, can you accept this PR, or we have to do some changes?

Sorry, things have been super busy here, so I've not had much time to delve into this. I'm a little hesitant to add the `file` method, because changes have multiple files, and just `file` is ambiguous. ISTM that we could instead add `deploy_file` in `plan.pm` at line 233, even if we keep `%F` as an alias for it. What do you think?

Looking at the code, I also suspect that the hash created at line 233 of `plan.pm` could be replaced with something else that defaulted to all the properties of a Change object, not just those we thought to add to this hash. Not asking for that here, though.

ghoust-xxx:
> I'm a little hesitant to add the `file` method, because changes have multiple files, and just `file` is ambiguous. ISTM that we could instead add `deploy_file` in `plan.pm` at line 233, even if we keep `%F` as an alias for it. What do you think?

Yes, you are right. Since we do not use `revert_dir` we can just use `deploy_file` everywhere. Updated.

> Looking at the code, I also suspect that the hash created at line 233 of `plan.pm` could be replaced with something else that defaulted to all the properties of a Change object, not just those we thought to add to this hash.

I think it will be a good idea if we can use the same hash in `plan.pm` as well as in `log.pm`. In this case formatting options can be the same.

theory (approved):
Looks great, thank you! Will get it merged shortly.

theory:
Would you mind squashing this down to a single commit that includes an item in the Changes file?

ghoust-xxx:
Sure. Here it is.

Thanks for reviewing!

theory (approved):

theory:
Awesome, thank you!

PR #762, opened 2023-06-04
Record search path failures on connect

### LINKED ISSUE #668: [Oracle] Sqitch does not detect missing schema
Steps to reproduce:
1. Run sqitch deploy --registry xxx, where xxx is a schema that does not exist on the target database; and where there is a sqitch registry in the connecting user's schema.

Expected behavior:
* Sqitch would report an error and take no action.

Actual behavior:
* Sqitch will query the registry in the connecting user's schema instead.

theory:
Instead of ignoring them, record when the schema can't be added to the search path. Then use that value to avoid querying the database in `_sync_plan`. This prevents Sqitch from finding a valid project record when the search path contains a valid changes table and record for the project, which can happen when one starts using the current schema for the registry (the default on Oracle) and then tries to use a new registry. Resolves #668.

To best make use of the new `_no_registry` attribute of Engine, rename the `initialized()` method to `_initialized()` and add a new `initialized()` method to check it before calling the engine-specific `_initialized()` method. Also rename the `initialize()` method to `_initialize()` and provide an `initialize()` implementation that sets `_no_registry` to false after a successful call to the engine-specific `_initialize()` method.

Update the DBI `connected` callbacks to return anytime `$dbh->do` returns false, thereby always propagating errors. Previously errors were ignored in MySQL and Snowflake (and we had an invalid warehouse name in the tests that needed fixing!).

For search path errors, call the new DBIEngine role `_handle_no_registry` method, which sets `_no_registry` to true and discards the error so that the DBI won't throw it once the callback function returns. Since the error handling is boolean-based, here, and the error handler is never called from inside the callback, remove the `try {}` block. It wasn't doing anything.

While at it, require DBI 1.631 to ensure consistent behavior of error handling in callbacks and to eliminate a workaround for older versions in the mysql engine.


lib/App/Sqitch/Engine/pg.pm
```diff
@@ -137,14 +137,11 @@ has dbh => (
...
-                    try {
-                        $dbh->do(
-                            'SET search_path = ?',
-                            undef, $self->registry
-                        );
-                        # https://www.nntp.perl.org/group/perl.dbi.dev/2013/11/msg7622.html
-                        $dbh->set_err(undef, undef) if $dbh->err;
-                    };
+                    $dbh->do(
+                        'SET search_path = ?',
```
vectro, line 141: I think in postgresql it is not an error to set the search_path to reference a nonexistent schema.

theory: Yes, right, that's fine — better even, because then the query of the `changes` table will fail:

```
david=# create table changes (id int);
CREATE TABLE

david=# set search_path to 'xxx';
SET

david=# insert into foo values(1);
ERROR:  relation "foo" does not exist
LINE 1: insert into foo values(1);
                    ^
```

This is just what we want.

theory: Added a comment in c6ccc306 about this.

vectro (approved):

PR #765, opened 2023-06-25
Use POSIX collation in the Postgres registry

### LINKED ISSUE #763: Provide option to specify the charset or collation of text columns in the registry tables
PostgreSQL [indicates](https://www.postgresql.org/docs/current/locale.html#id-1.6.11.3.5):

> The drawback of using locales other than C or POSIX in PostgreSQL is its performance impact. It slows character handling and prevents ordinary indexes from being used by LIKE. For this reason use locales only if you actually need them.

Could Sqitch create its tables using the C collation rather than the default collation?

theory:
Applies also to Yugabyte, and separately add for Cockroach. Using the POSIX collation ensures that index ordering never changes when the database is upgraded, which is a particular problem with glibc collations, but since POSIX/C collation is strictly byte-ordered, it should be fine. Of course, any of use of `ORDER BY` on such columns will return unexpected results when users are used to other locales, but since Sqitch itself only ever orders by timestamp, it should not be an issue in its own use.

Closes #763.

### DIFF
--- a/Changes
+++ b/Changes
@@ -49,6 +49,9 @@ Revision history for Perl extension App::Sqitch
        its Unicode improvements. Thanks to Mark Tyrrell for the report and
        @tiberiusferreira and Perl Monks `1nickt` and`InfiniteSilence` for the
        feedback (#825).
+     - Set all text column collations to "POSIX" on Postgres 9.1 and higher.
+       and Yugabyte 2.9 and higher. Does not apply to existing registries,
+       only new ones. Thanks to @datafoo for the suggestion (#763)!
 
 1.4.1  2024-02-04T16:35:32Z
      - Removed the quoting of the role and warehouse identifiers that was
--- a/t/pg.t
+++ b/t/pg.t
@@ -371,6 +371,7 @@ RUNREG: {
     $mock_engine->mock(_psql_major_version => sub { $psql_maj });
     my @ran;
     $mock_engine->mock(_run => sub { shift; push @ran, \@_ });
+    $mock_engine->mock(_provider => 'postgres');
 
     # Mock up the database handle.
     my $dbh = DBI->connect('dbi:Mem:', undef, undef, {});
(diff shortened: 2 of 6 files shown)


theory:
> As I would always have to reindex all indexes in a database with reindexdb, I don't see any advantage.

That's what I was wondering about.

> In addition outside of certain countries (US, CA, UK) email addresses with characters outside of the POSIX character set are possible and in fact used.

I believe you can still use UTF-8 bytes, it's just that things are indexed purely in byte order, not a character encoding order. I *think* that's right, but I'm not positive, admittedly.

theory:
The more I think about it, the more convinced I am that this isn't the right thing to do. We shouldn't be making collation decisions for the people using the database. It might even be surprising that the registry tables use a different collation than the rest of the database. And as @sjstoelting points out, one ought to expect to reindex all indexes in a database where the collation has changed; and including Sqitch's indexes should be pretty obvious (and generally low-overhead, it's not usually much data).

What do you think, @datafoo?

theory:
> Is there any reason to use locales?

The thing is, we don't. We just default to whatever the DBA has configured things for. It might be more surprising to them if it _wasn't_ the same.

sjstoelting:
And again, there's a world outside the USA which do need and use different collations.
If you don't take the characters used in other countries seriously, I boycott your products wherever possible.
Leave it to the users and don't make decisions, that don't fit elsewhere.

And again, I even have email addresses with domains containing umlauts.

sjstoelting:
What is the goal about having data stored as UTF-8 and sorted by something that doesn't contain most of the characters that are available?

You can change your own stuff to whatever you like, if you love POSIX, do it in your own stuff.
But dictating that to everyone is still something I believe to be a decision by it's users.

Am 15. August 2023 11:52:37 MESZ schrieb datafoo ***@***.***>:

theory:
I'm beginning to come back around to @datafoo's point of view, because using C/POSIX collation has no effect on the data you can store: it's all UTF-8. What it affects is the sort order:


```
david=# CREATE TABLE try (name TEXT COLLATE "POSIX", email TEXT COLLATE "POSIX");
CREATE TABLE
david=# INSERT INTO try VALUES('Stölting', 'stölting@example.com'), ('Stõlting', 'stõlting@example.com'), ('Stolting', 'stolting@example.com');
INSERT 0 3
david=# select * from try order by name;
   name   |        email         
----------+----------------------
 Stolting | stolting@example.com
 Stõlting | stõlting@example.com
 Stölting | stölting@example.com
(3 rows)
```

All the character encodings remain, although the sort ordering might not be expected for some locale or another: it's simply sorted in lexicographic order — that is, by the UTF-8 byte orders.

Changes
```diff
@@ -30,6 +30,9 @@ Revision history for Perl extension App::Sqitch
        Exasol and because `FAILURE` maps to exit code `1`, which has in the
        past been more akin to a warning from Sqitch. Thanks to @vectro for the
        report (#831).
+     - Set all text column collations to "POSIX" on Postgres 9.1 and higher.
+       and Yugabyte 2.9 and higher. Does not apply to existing registries,
+       only new ones. Thanks to @datafoo for the suggestion (#763)!
```
autarch, line 35: Doesn't this also apply when you upgrade the registry, at least for some columns?

t/pg.t
```diff
@@ -437,14 +438,46 @@ RUNREG: {
...
-    is_deeply \@sra_args, [], 'Still hould not have have called selectrow_array';
+    is_deeply \@sra_args, [], 'Still should not have have called selectrow_array';
     is_deeply \@done, [['SET search_path = ?', undef, $registry]],
         'The registry should have been added to the search path again';
 
     # Make sure the file was changed to remove SCHEMA IF NOT EXISTS.
     file_contents_like $tmp_fh, qr/\QCREATE SCHEMA :"registry";/,
         'Should have removed IF NOT EXISTS from CREATE SCHEMA';
 
+    # Make sure we havne't removed collation.
```
autarch, line 449:
```suggestion
    # Make sure we haven't removed collation.
```

autarch (approved):
LGTM % a few questions and one typo.

Changes
theory: It would if we added more upgrades, that's true.

theory:
I was close to merging this PR, but then three things happened:

* @autarch pointed out the upgrade challenges in a comment

* @ardentperf pointed out via Slack that stuff like `initcap()` does not work as expected with POSIX collation:

    ``` psql
    postgres=# select initcap('stõlting' collate "C.UTF-8");
    initcap
    ----------
    Stõlting
    (1 row)
```

* I realized that the same issue exists for the other database engines, so shouldn't they get the same treatment?

To this last point, I spent a few hours experimenting with just setting the URI and SHA hash hex string columns to a relevant encoding or collation for all the engines. The results are in #857. It's not a super promising start. More notes there, but I think it'd make sense to pick up the work there, rather than here, should there be energy for or demand to let users determine the character set or encoding to use for text columns.

In the meantime, I'll update #763 with a workaround for Postgres.

PR #772, opened 2023-07-26
Fix cockroachdb

NOBLES5E:
Since cockroach is not listed in https://github.com/sqitchers/sqitch/blob/v1.3.1/lib/App/Sqitch/Command.pm#L17, there's no way to specify engine as `cockroach`.


theory:
Thank you, can't believe I missed this. Closing in favor #774, which also adds a test to ensure it doesn't happen again.

PR #773, opened 2023-07-29
Remove remaining uses of the smartmatch operator

theory:
Most were removed in 2013, but with Perl 5.38 formally deprecating them (and issuing noisy warnings), it seems prudent to simply replace them with more traditional Perl patterns like the ternary operator and regular expressions.

### DIFF
--- a/lib/App/Sqitch/ItemFormatter.pm
+++ b/lib/App/Sqitch/ItemFormatter.pm
@@ -50,51 +50,46 @@ has formatter => (
     isa     => class_type('String::Formatter'),
     default => sub {
         my $self = shift;
-        no if $] >= 5.017011, warnings => 'experimental::smartmatch';
         String::Formatter->new({
             input_processor => 'require_single_input',
             string_replacer => 'method_replace',
             codes => {
                 e => sub { $_[0]->{event} },
                 L => sub {
-                    given ($_[0]->{event}) {
-                        when ('deploy') { return __ 'Deploy' }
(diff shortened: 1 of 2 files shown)


lib/App/Sqitch/ItemFormatter.pm
```diff
@@ -110,8 +105,8 @@ has formatter => (
                 c => sub {
                     return "$_[0]->{committer_name} <$_[0]->{committer_email}>"
                         unless defined $_[1];
-                    return $_[0]->{committer_name}  if $_[1] ~~ [qw(n name)];
-                    return $_[0]->{committer_email} if $_[1] ~~ [qw(e email)];
+                    return $_[0]->{committer_name}  if $_[1] =~ /^n(?:ame)?/;
```
autarch, line 108: Technically `$` isn't quite the same since it allows for an optional newline at the end of the string, so it'd match the literal `"name\n"`. To actually match the end of the string you need `\z`. In that case you might also want to use `\A` at the beginning for symmetry, but without the `/m` flag this isn't necessary.

autarch (changes requested):
LGTM except for one regex thing. It's a bit picky but I think it's best to be explicit here.

lib/App/Sqitch/ItemFormatter.pm
theory: Can't believe I forgot the `$`. Changed to `\A` and `\z` in 46b8d78.

PR #774, opened 2023-07-29
Add cockroach to Command's list of engines

theory:
The list recognizes engine names on the command-line. Add a test to ensure new engines added in the future won't be omitted. Thanks to @NOBLES5E for the spot!

### DIFF
--- a/t/command.t
+++ b/t/command.t
@@ -4,7 +4,7 @@ use strict;
 use warnings;
 use 5.010;
 use utf8;
-use Test::More tests => 198;
+use Test::More tests => 199;
 #use Test::More 'no_plan';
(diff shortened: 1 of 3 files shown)


t/command.t
```diff
@@ -131,6 +131,15 @@ DEBUG: {
         'Should have sent error to debug';
 }
 
+##############################################################################
+# Test ENGINES.s
```
autarch, line 135: That last "s" is a typo, right?

autarch (approved):
LGTM % one small typo.

t/command.t
theory: Yup, thanks.

PR #785, opened 2023-09-02
Fix output verbosity for changes to be reverted

### LINKED ISSUE #784: Changes to be reverted are printed with `core.verbosity = 1`
The release notes for 1.4.0 say:

> Output the list of changes to be deployed or reverted when `--verbose` is specified at least twice. Thanks to @vectro for the PR (https://github.com/sqitchers/sqitch/pull/702).

But the changes to be reverted are printed even with `core.verbosity = 1`. The changes to be deployed are correctly printed with `core.verbosity = 2` only.

The revert case is logged with `info` instead of `debug`:

https://github.com/sqitchers/sqitch/blob/fd4b1901d1414ae62f4a2aea8115c5cebc4136f3/lib/App/Sqitch/Engine.pm#L373

ewie:
Commit 06daffc4 changed the output of deploy and revert to also print the changes to be deployed or reverted.  Changes to be deployed are printed for verbosity > 1.  But changes to be reverted are printed for verbosity > 0.  Fix that and require verbosity > 1 in both cases.

Fixes #784.

### DIFF
--- a/lib/App/Sqitch/Engine.pm
+++ b/lib/App/Sqitch/Engine.pm
@@ -370,11 +370,11 @@ sub revert {
                 change      => $change->format_name_with_tags,
                 destination => $self->destination,
             ));
-            $sqitch->info(__ 'Will revert the following changes:');
-            map { $sqitch->info($_) } @change_descriptions;
+            $sqitch->debug(__ 'Will revert the following changes:');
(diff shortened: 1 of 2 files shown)


theory:
Thank you! Merged in 7d56f106.

PR #792, opened 2023-10-02
Handle macOS mariadb client

### LINKED ISSUE #791: Fails to build on macOS Ventura when using MariaDB
Hi,

When using brew package `mariadb` in macOS Sqitch fails to compile. For some reason the result of `mysql --version` is slightly different from Linux and probably other Unixes.

Here is the output of `mysql --version` on Debian 12

    mysql  Ver 15.1 Distrib 10.11.3-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper

Here is the output of `mysql --version` on macOS Ventura

    mysql from 11.1.2-MariaDB, client 15.2 for osx10.18 (x86_64) using  EditLine wrapper

Due to this difference [this](https://github.com/sqitchers/sqitch/blob/v1.4.0/t/mysql.t#L78) line returns `undef` and the same problem can be found [here](https://github.com/sqitchers/sqitch/blob/v1.4.0/lib/App/Sqitch/Engine/mysql.pm#L182).

I might be able to provide a PR for this.

carragom:
Fix #791

### DIFF
--- a/lib/App/Sqitch/Engine/mysql.pm
+++ b/lib/App/Sqitch/Engine/mysql.pm
@@ -179,7 +179,7 @@ has _mysql => (
         # Get Maria to abort properly on error.
         my $vinfo = try { $self->sqitch->probe($self->client, '--version') } || '';
         if ($vinfo =~ /mariadb/i) {
-            my ($version) = $vinfo =~ /Ver\s(\S+)/;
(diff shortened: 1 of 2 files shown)


carragom:
Any idea why this is failing ?. Snowflake is reporting connection errors and Windows errors are native code related.

theory:
Yeah, it's a permissions issue, don't sweat it.

theory (approved):

PR #832, opened 2024-06-06
Ban "[" and "]" in plan names

### LINKED ISSUE #829: The change names containing `]` are not parsed correctly when used as dependencies
For example:
```
%syntax-version=1.0.0
%project=test

testa[test]test 2024-05-23T11:14:35Z Žiga Leber <mail+github@zigaleber.com> # Add schema
testb [testa[test]test] 2024-05-23T11:16:00Z Žiga Leber <mail+github@zigaleber.com> # Something
```

This fails with:
```console
Syntax error in sqitch.plan at line 5: Invalid name; names must not begin with punctuation, contain "@", ":", "#", "\", or blanks, or end in punctuation or digits following punctuation
```

The issue arises due [to](https://github.com/sqitchers/sqitch/blob/5522821bdee36523bc9ab36d7c550834ca2c2748/lib/App/Sqitch/Plan.pm#L301), the regex assumes the dependencies cannot contain `]`. The example is then parsed as `testb [testa[test]` followed by `test]`.

The easiest solution would be to revise the rule to: All other characters may be any UTF-8 character other than “:”, “@”, “#”, “\” and "]".

theory:
Because they can mess with dependency parsing, which always disallowed brackets because the wrap the list of dependencies. Fixes #829.


autarch (approved):
LGTM!

theory:
Thanks @autarch, always glad to have someone look things over.

PR #847, opened 2024-12-30
Properly avoid advisory Yugabyte lock errors

### LINKED ISSUE #841: Sqitch no longer detects Yugabyte
From comments on #635, with debugging by @stefanoric, it appears that Sqitch is failing to recognize Yugabyte for some reason. Most likely something has changed such that the detection code no longer works:

https://github.com/sqitchers/sqitch/blob/a7dfdb3b16979bf9cdd414af10eb1a7236a36066/lib/App/Sqitch/Engine/pg.pm#L147-L152

### LINKED ISSUE #845: Sqitch with Oracle DB doesn't seem to properly respect SQITCH_URI
**tl;dr**
Attempted to use docker-sqitch to avoid having to disable SIP. Continued to get "ORA-12162: TNS:net service name is incorrectly specified" and finally tracked it down to `SQITCH_URI` not being respected - or something.

**Reproduction Steps**
1. Build sqitch:sqitch/oracle image
2. Create ~/.sqlplus/tnsnames.ora
3. Launch OracleDB in k8s cluster with a service type `LoadBalancer`
```
TESTDB =
  (DESCRIPTION =
    (ADDRESS = (PROTOCOL = TCP)(HOST = 10.0.10.3)(PORT = 1521))
    (CONNECT_DATA =
      (SERVER = DEDICATED)
      (SERVICE_NAME = freepdb1)
    )
  )
```
4. `export SQITCH_URI=db:oracle://user:password@localhost:1521/testdb`
5. `export SQITCH_IMAGE=`sqitch/sqitch:oracle`
6. `export TNS_ADMIN=/home/.sqlplus`
7. `docker-sqitch.sh status`
```
# On database db:oracle:
ORA-12162: TNS:net service name is incorrectly specified
Help: https://docs.oracle.com/error-help/db/ora-12162/ (DBD ERROR: OCIServerAttach)
```
9. `sqitch target add testdb db:oracle://user:password@localhost:1521/testdb`
8. `docker-sqitch.sh status -t testdb`
```
# On database ocadb
No changes deployed
```

Was able to track this down by verifying connection in `/instantclient/sqlplus` and then trying a bunch of things until realizing manually specifying the target (either via name or URI) worked.

theory:
The `try_lock` method was not properly checking for Yugabyte before attempting to acquire an advisory lock. Likely it used to be a warning, but now is an error that cannot be avoided. So have `try_lock` check for Yugabyte before attempting to take an advisory lock. This requires fetching the database handle, first, as the database is not identified until it connects. So abstract that need by replacing the `_provider` attribute with a method that checks a DBI private module attribute. Fixes #841.

While at it, replace a wayward mention of the long-unsupported (or maybe incorrectly documented?) `SQITCH_URI` environment variable in the Oracle tutorial with the correct name, `SQITCH_TARGET`. Resolves #845.

### DIFF
--- a/lib/sqitchtutorial-oracle.pod
+++ b/lib/sqitchtutorial-oracle.pod
@@ -306,10 +306,10 @@ using the Docker or VM environments described above, you might need to
 L<create the database|https://docs.oracle.com/cd/B28359_01/server.111/b28310/create001.htm#ADMIN11068>
 and configure the SID. Assuming you have an Oracle SID named C<FLIPR_TEST> set
 up in your C<F<TNSNAMES.ORA>|https://www.orafaq.com/wiki/Tnsnames.ora> file,
-tell Sqitch where to send the change via a
+tell Sqitch where to send the change to a target specified as a
 L<database URI|https://github.com/libwww-perl/uri-db/>, such as
 
-  export SQITCH_URI=db:oracle://$username:$password@/flipr_test
+  export SQITCH_TARGET=db:oracle://$username:$password@/flipr_test
 
 With that URI set up, we can deploy:
 
--- a/Changes
+++ b/Changes
@@ -10,6 +10,13 @@ Revision history for Perl extension App::Sqitch
        exception objects rather than strings, for more consistent error
        handling throughout.
      - Removed duplicate DBI error handling code from engines and tests.
+     - Fixed an order of operation issue that prevented Sqitch from detecting
+       Yugabyte before attempting to create an advisory lock, which results in
+       an error for more recent Yugabyte releases. Thanks to Stefano Ricciardi
+       for the report (#841).
+     - Removed a wayward mention of the long-deprecated `SQITCH_URI`
+       environment variable from the Oracle tutorial. Thanks to Austin Hanson
+       for the report (#845).
 
 1.4.1  2024-02-04T16:35:32Z
      - Removed the quoting of the role and warehouse identifiers that was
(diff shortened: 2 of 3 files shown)


theory:
@FranckPachot how far back does Yugabyte respect the `yb_silence_advisory_locks_not_supported_error` GUC?

I don't suppose there is also a GUC to suppress errors about the use of `LOCK TABLE changes IN EXCLUSIVE MODE`?

lib/sqitchtutorial-oracle.pod
autarch, line 1: Did you mean to change this file as part of this PR?

autarch (changes requested):
The PR description mentions setting `yb_silence_advisory_locks_not_supported_error` but I don't see that anywhere in this PR. Also, there's a change for the Oracle docs that maybe belongs in a different PR?

lib/sqitchtutorial-oracle.pod
theory: Yes, this was part of the fix for #845. Was not part of the original PR, but rolled in a later rebase and forgot to update the PR description.

theory:
I removed the `yb_silence_advisory_locks_not_supported_error` bit because it failed on earlier versions of Yugabyte and I couldn't figure out how to suppress the errors inside the `connected` callback.

Much obliged for the review, @autarch!

autarch (approved):
LGTM!

FranckPachot:
> @FranckPachot how far back does Yugabyte respect the `yb_silence_advisory_locks_not_supported_error` GUC?
> 
> I don't suppose there is also a GUC to suppress errors about the use of `LOCK TABLE changes IN EXCLUSIVE MODE`?

This GUC is only about advisory lock. There's no GUC to ignore LOCK TABLE.
It is much better to avoid calling locks rather than having them ignore ( personally, I don't like this GUC - ignoring unsupported feature)

theory:
Thanks, @FranckPachot. I ended up removing it anyway.

PR #849, opened 2024-12-31
Exit `sqlplus` on error with error code 4

### LINKED ISSUE #831: WHENEVER SQLERROR EXIT SQL.SQLCODE is not safe on UNIX
On Oracle, Sqitch runs the SQL\*Plus command `WHENEVER SQLERROR EXIT SQL.SQLCODE`, so that we can detect errors. However, on UNIX, SQL\*Plus will return the actual error code modulo 256, which means the application can return zero exit status if the error code is a multiple of 256.

For example, if Sqitch tries to deploy a script creating a table with too many columns, causing error ORA-01792: maximum number of columns in a table or view is 1000, SQL*Plus will still exit with zero exit status (because 1792 mod 256 = 0), so Sqitch will not detect the error.

theory:
Change the default error code from Oracle `sqlplus` from `SQL.SQLCODE` to `4`, because otherwise the exit code is returned `modulo 256`, meaning it can end up `0` for success for an SQL error code like `ORA-01792`. Selected `4` as the replacement to match the behavior of Exasol and because `FAILURE` maps to exit code `1`, which has in the past been more akin to a warning from Sqitch. Resolves #831.

### DIFF
--- a/Changes
+++ b/Changes
@@ -23,6 +23,13 @@ Revision history for Perl extension App::Sqitch
        underlying error loading the engine package and the stack trace remain
        available by using triple verbosity (`-vvv`). Thanks to Martin Fischer
        for the report (#838).
+     - Changed the default error code from Oracle `sqlplus` from `SQL.SQLCODE`
+       to `4`, because otherwise the exit code is returned `modulo 256`,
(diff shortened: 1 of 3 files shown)


autarch (approved):
LGTM!

PR #850, opened 2025-01-01
Always detect missing deploy and revert scsripts

theory:
Add checks for the existence of deploy and revert scripts before executing them (or logging them). Requires the presence of various test files in order for `t/engine.t` to work, as it previously created new changes willy-nilly without needing or caring about deploy and revert scripts. Hence all the new files in `t/sql`.


theory:
Merged in 7388a135bc8088b29312cacce4c47ea09abc77aa.

PR #851, opened 2025-01-01
Properly handle MySQL URIs without a DB name

### LINKED ISSUE #821: `sqitch deploy` requires the `uri` to point to an existing schema, making it hard to start from an empty state
I'm setting up a sqitch project where I want to include the initial schema creation among the migration files:

`deploy/schemas/appschema.sql`

```sql
-- Deploy app:schemas/appschema to mysql
begin;

CREATE DATABASE IF NOT EXISTS appschema;

COMMIT;
```

I therefore have a `sqitch.conf` with

```
[target "dev"]
	uri = db:mysql://root:password@127.0.0.1:3306/
	registry = appschema_sqitch
```

However, this causes `sqitch deploy` to fail on a clean database setup (mysql container):

```sh
$ sqitch deploy
warning: Database name missing in URI "db:mysql://root:password@127.0.0.1:3306/"
"appschema_sqitch" failed to start: "No such file or directory
```

**Expected behavior**

URIs without a target database should be accepted. It should be up to the user to qualify their statements or call e.g. `USE appschema` in scripts in such cases.

---

Sqitch: v1.4.1
OS: Debian 11 (running in devcontainer under Ubuntu 20.04.6 LTS (focal) (64-bit) on Docker version 25.0.4, build 1a576c5)
Database: MySql 8.0.32

theory:
The MySQL engine has in theory always allowed a URI without a database name, though it emits a warning. But the `mysql` CLI option handling did not properly account for it, expecting to always find a `--database` option to replace.

Fix it by replacing the `--database` option when it exists and appending it when it doesn't. Fixes #821.

### DIFF
--- a/Changes
+++ b/Changes
@@ -34,6 +34,9 @@ Revision history for Perl extension App::Sqitch
        deploying and reverting. Previously Sqitch deferred such errors to the
        CLIs, but they're never called when using `--log-only`. Thanks to
        @vectro and Erik Wienhold for the suggestion (#828).
+     - Fixed a bug where the MySQL engine failed to properly handle target
+       URIs with no database name. Thanks to Felix Zedén Yverås for the report
+       (#821).
 
 1.4.1  2024-02-04T16:35:32Z
      - Removed the quoting of the role and warehouse identifiers that was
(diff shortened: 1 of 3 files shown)


theory:
Merged in f3963270bbde6d24c602b251c758a85edd89c11d.

PR #852, opened 2025-01-02
Omit MySQL checkit() when it cannot be created

### LINKED ISSUE #824: Checkit function creation causing error when binary logging enabled - MySQL DB
We have binary logging enabled on a MySQL 5.7 DB but the deploy scripts fail with SQL Error 1419 

"You do not have super user privilege and binary logging is enabled"

This error is caused by the creation of the checkit function, however we do not use the function in our verify scripts.

Is there a fix for this that doesn't involve setting log_bin_trust_function_creators to true?

theory:
Remove the use of the `checkit()` function from the MySQL triggers that enforce constraints on the `dependencies` table, then modify the deployment of MySQL registry scripts to remove `checkit()` when MySQL is using binary logging and Sqitch is not a super user. Fixes #824, which reports that MySQL with binary logging enabled disallows the creation of `DETERMINISTIC` functions by non-super users. In such a configuration, Sqitch itself functions exactly as before, but `checkit()` is not available for use in verify scripts.

Refactor the MySQL engine's `run_upgrade()` method to move the rejiggering of a file to a new method, `_prepare_registry_file`. Teach that method then to remove bits of the script as appropriate to the MySQL version and permissions.

Update the MySQL tutorial to note how to add `checkit()` if it's missing.

### DIFF
--- a/Changes
+++ b/Changes
@@ -37,6 +37,9 @@ Revision history for Perl extension App::Sqitch
      - Fixed a bug where the MySQL engine failed to properly handle target
        URIs with no database name. Thanks to Felix Zedén Yverås for the report
        (#821).
+     - Updated the MySQL engine to omit the `checkit()` function when using
+       binary logging and Sqitch lacks super user permissions. Thanks to Scott
+       Edwards for the report and to Janosch Peters for the solution (#824).
 
 1.4.1  2024-02-04T16:35:32Z
      - Removed the quoting of the role and warehouse identifiers that was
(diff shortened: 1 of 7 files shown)


theory:
I'm not positive this is the best way to detect in a single query whether the current user is a super user:

```sql
SELECT super_priv = 'Y' FROM mysql.user
 WHERE CONCAT(user, '@', host) = current_user
```

It works, but I don't know what all the wrinkles are. Like, are there conditions when `current_user` quotes the username or hostname?

autarch (approved):
LGTM!

theory:
Merged in 93a0d1f867605fed967e70d7fe4c1d90ef1cc57b.

PR #853, opened 2025-01-03
Skip registry creation on Snowflake when necessary

### LINKED ISSUE #828: Unexpected behavior when deploy script is missing
## Steps to reproduce
1. Set up a Sqitch project where some changes’ deploy scripts are missing.
1. Run `sqitch deploy`

Or equivalently, run these commands:
```
sqitch init ...
sqitch add test1
sqitch add test2
rm deploy/test1.sql
rm deploy/test2.sql
sqitch deploy
```

## Expected behavior

- Sqitch should fail with an error the first time a plan with a missing deploy script is encountered, preferably before deploying anything.

## Actual behavior

- The first change with a missing deploy script succeeds, leaving a null value in changes.script_hash.
- The second change with a missing deploy script fails due to the unique constraint on changes.script_hash, with the confusing error message “The deploy script is not unique”.

## Other notes
Tested using Sqitch 1.4.1 with the Oracle database driver.
On Oracle (and some other platforms), a unique constraint on a null column allows at most one null value on that column.

theory:
In some cases, the Sqitch user will lack permission to create a schema, such as when the DBA doesn't want to give it that permission. Implement `run_upgrade` in the Snowflake engine to detect this lack of permission by executing `CREATE SCHEMA IF NOT EXISTS` and stripping out `CREATE SCHEMA` and `COMMENT ON SCHEMA` commands when it fails due to insufficient privileges. Resolves #828.

### DIFF
--- a/Changes
+++ b/Changes
@@ -40,6 +40,10 @@ Revision history for Perl extension App::Sqitch
      - Updated the MySQL engine to omit the `checkit()` function when using
        binary logging and Sqitch lacks super user permissions. Thanks to Scott
        Edwards for the report and to Janosch Peters for the solution (#824).
+     - Taught the Snowflake engine to detect when the Sqitch user lacks
+       permission to create a schema and to skip the creation of the registry
+       schema. Useful for cases when the registry schema was created in
+       advance. Thanks to Peter Wimsey for the suggestion (#826).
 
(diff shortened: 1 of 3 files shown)


autarch (approved):
LGTM!

PR #854, opened 2025-01-03
Switch from DBD::mysql to DBD::MariaDB

### LINKED ISSUE #825: Support DBD::MariaDB
Since version 5 (Oct 2023), DBD::mysql cannot be installed without MySQL 8.x. This ranges from inconvenient to unacceptable for applications using MariaDB (particularly considering that the the necessary mariadb libs are packaged with the official perl docker images, while mysql's are not). Fortunately, since the release of [DBIx::Class::Storage::DBI::MariaDB](https://metacpan.org/pod/DBIx::Class::Storage::DBI::MariaDB) in August 2023, [DBD::MariaDB](https://metacpan.org/dist/DBD-MariaDB/view/lib/DBD/MariaDB.pod) is a viable option for most applications as a drop-in replacement for DBD::mysql.

Unfortunately, Sqitch is hardcoded to use the DBD::mysql driver for MariaDB databases, which requires the installation of DBD::mysql 5.x (and consequently MySQL 8.x), or the no-longer-maintained/6-year-old DBD::mysql 4.x. Would you consider adding support for DBD::MariaDB as a much cleaner solution?

theory:
DBD::mysql 5 recently dropped support for building from MySQL client libraries prior to v8. This creates challenges for people still using earlier versions, as well as those using MariaDB.

DBD::MariaDB forked from DBD::mysql back in 2018, and maintains compatibility with both MySQL and MariaDB. It remains compatible with earlier versions of MySQL as well as the MariaDB client libraries. It also includes bug fixes and more transparent support for Unicode encoding.

So switch from DBD::mysql to DBD::MariaDB. It's largely a drop-in replacement, with just a few additional changes:

*   Add `_dsn()` to DBIEngine and always use it to connect to the database. Override it in mysql.pm to replace `dbi:mysql` with `dbi:MariaDB`. Also override it in some other drivers that mangle the value returned by URI::db.
*   Remove incompatible config parameters `mysql_enable_utf8` (no longer needed due to the aforementioned Unicode bug fixes), `mysql_auto_reconnect`, and `mysql_use_result`.
*   Read DBI handle attributes starting with `mariadb_` instead of `mysql_`.
*   Add a `USE` statement to set the registry as the current database after it has been created. Not sure how this wasn't required when using DBD::mysql, but aligns more closely with the requirements of other engines.

Resolves #825.


autarch (approved):
LGTM!

theory:
Merged in d17dff97.

PR #863, opened 2025-01-18
Restore MySQL connection to the registry

### LINKED ISSUE #862: MariaDB: Sqitch doesn't respect registry config anymore
Hello, previous to https://github.com/sqitchers/sqitch/pull/854 sqitch used to respect the registry and use that value as the sqitch schema.

That doesn't seem to be the case anymore. Maybe that was a MySQL specific behavior? If so, could it be replicated for MariaDB?

Example of error:

```
# On database TestMariaDB
Table 'TestSchema.changes' doesn't exist
```

With config:
```
[core]
	engine = mariadb
	registry = sqitch
```

theory:
A bug introduced in d17dff9 and shipped in v1.5.0 caused the MySQL engine to connect to the target database instead of the registry database. Restore connecting to the registry database. Resolves #862.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,9 @@
 Revision history for Perl extension App::Sqitch
 
-1.5.1  Note yet released
+1.5.1  Not yet released
+     - Fixed a bug introduced in v1.5.0 where the MySQL engine connected to
+       the target database instead of the registry database. Thanks to
+       @tiberiusferreira for the report (#862)!
 
 1.5.0  2025-01-08T03:22:40
      - Fix improperly nested Pod headers that were incrementing two levels
--- a/lib/App/Sqitch/Engine/mysql.pm
+++ b/lib/App/Sqitch/Engine/mysql.pm
@@ -65,7 +65,7 @@ sub _def_user { $_[0]->_mycnf->{user} || $_[0]->sqitch->sysuser }
 sub _def_pass { $ENV{MYSQL_PWD} || shift->_mycnf->{password} }
 
 sub _dsn {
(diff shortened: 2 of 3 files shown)


tiberiusferreira:
Hello, is there an easy way to test this with brew? 
I'm trying to compile from source and installed Dist::Zilla:
```
cpanm Dist::Zilla
Dist::Zilla is up to date. (6.032)
```
But it didn't seem to install dzil in the correct path:
```
tiberiodarferreira@Tiberios-MacBook-Pro sqitch % dzil
zsh: command not found: dzil
```

theory:
Oh, yikes, don't do that! Easiest would probably be to just edit the file. The file should live here:

``` console
❯ ls -l "$(brew --prefix)"/Cellar/sqitch/1.5.0/lib/perl5/App/Sqitch/Engine/mysql.pm 

-r--r--r--  1 david  admin  19113 Jan  8 07:50 /opt/homebrew/Cellar/sqitch/1.5.0/lib/perl5/App/Sqitch/Engine/mysql.pm
```

If so, then you can edit it with these commands:

``` sh
chmod +w "$(brew --prefix)"/Cellar/sqitch/1.5.0/lib/perl5/App/Sqitch/Engine/mysql.pm
perl -i -pe 's/shift->uri/shift->registry_uri/' "$(brew --prefix)"/Cellar/sqitch/1.5.0/lib/perl5/App/Sqitch/Engine/mysql.pm
```

Then check to make sure the line was updated properly. Should be good go go then.

tiberiusferreira:
Oh, I totally missed the idea of just editing the installed file directly! 
It worked, tested it and the fix works!

theory:
Thank you!

PR #873, opened 2025-03-16
chore: added missing german translation

0xflotus:


theory:
Thanks! There are a couple more that are missing. Care to update them, too?

```
msgid "File {file} has a double extension of {ext}"
msgid "Must specify a target revision in strict mode"
msgid "Will deploy the following changes:"
msgid "Missing required parameter $prompt_default"
msgid "Will revert the following changes:"
msgid "Would revert the following changes:"
msgid ""
"Blocked by another instance of Sqitch working on {dest}; waiting {secs} "
"seconds..."

msgid ""
"Timed out waiting {secs} seconds for another instance of Sqitch to finish "
"work on {dest}"

msgid "Insufficient permissions to create the checkit() function; skipping."

msgid ""
"Because the \"changes\" table does not exist, Sqitch will now initialize the "
"database to create its registry tables."

msgid "Tag \"{tag}\" contains illegal character {sep}"

msgid ""
"\"{command}\" cannot be used in strict mode.\\nUse explicity revert and "
"deploy commands instead."
```

theory:
Thank you. Perhaps it's my ignorance of German, but are all those words really supposed to be title-cased?

0xflotus:
> Thank you. Perhaps it's my ignorance of German, but are all those words really supposed to be title-cased?

Nouns in general are title-cased in german. Yes... 😅

theory (approved):
Thank you!

Issue #839, opened 2024-11-19
`sqitch.conf` file permission mask


fluca1978:
Using `sqitch` `1.4.1` from CPAN.
I noted that `sqitch.conf` has default read permission for everyone:

```
% sqitch --version
sqitch (App::Sqitch) v1.4.1

 % ls -l sqitch.conf
-rw-rw-r-- 1 luca luca 222 nov 19 09:41 sqitch.conf
```

Apparently the application is working even if permissions are set for only the user, thus:

```
% chmod 600 sqitch.conf
```

Shouldn't this be default permission mask since `sqitch.conf` could possibly contain connection passwords?

theory:
Never thought about it. It defaults to whatever the current directory's umask is.

theory:
[Looks like](https://github.com/bestpractical/config-gitlike/blob/a065320af482d33142116f8ecf67606bdb9da48d/lib/Config/GitLike.pm#L1127C1-L1128C1) Config::GitLike, which creates the file, does not have any support for custom permissions:

```pl
    sysopen(my $fh, "${filename}.lock", O_CREAT|O_EXCL|O_WRONLY)
        or die "Can't open ${filename}.lock for writing: $!\n";
```

Looks like it would require another attribute on the object, similar to `encoding`, to specify permissions to pass as the fourth argument to [`sysopen`](https://perldoc.perl.org/functions/sysopen), something like:

```pl
sysopen(my $fh, "${filename}.lock", O_CREAT|O_EXCL|O_WRONLY, $self->permissions || 0666)
```

If they added such an option I'd be happy to use it in Sqitch.

fluca1978:
Would it have `sqitch` to automatically adjust permissions at first read  a possible solution?

fluca1978:
@theory  PTAL at <https://github.com/fluca1978/sqitch/commit/d4ab60e57d217220f10af1878d1f94b45febcbf2>. Seems very simple solution to me, but works for both adding and editing configuration.
Clearly it makes extra work, since it checks that the configuration file(s) have the right permissions every time they are loaded, but at least it adjusts the permissions as soon as possible.
Clearly, there is no solution against a user that deliberately taint his own environment changing the permissions outside of the `sqitch` application.

fluca1978:
> Config::GitLike operates on a whole [chain of config files](https://sqitch.org/docs/manual/sqitch-config/#files), including, potentially, files not owned by the current user. I think this should be limited to those owned by the current user, no?
> 

[...]

PR #901, opened 2025-09-01
[#839] Adjusts permissions of `sqitch.conf` (and friends) every time

fluca1978:
Every time the configuration is loaded, the file permissions are checked and if they are different from `0600`, they are adjusted via an explicit `chmod`.
The rationale is to ensure, as much as possibile, that configuration including target credentials remains private to the user.

This introduces the dependency with `File::stat`, in order to perform a more OOP-`stat` call.
The `App::Sqitch::Config::load_file` method has changed in order to call a new function, `::chmod_files_if_needed` that accepts a list of file-ish stuff to check against permissions.


theory:
I'm not sure this is desirable or even useful, as Stefanie Janine Stölting [points out](https://digitalcourage.social/@sjstoelting/115272127885656313):

> On git clone all file are becoming owned by the current user and the user is also the group. The attributes are derived by the ones from the repository.
>
> When you change the file rights with chmod  git shows the file as changed.

Git doesn't enforce something like this, AFAICT. I'm not sure it makes sense for Sqitch to do so, either.

PR #902, opened 2025-09-05
Add ClickHouse Engine

theory:
# 📢❗🚨New engine! New engine! New engine!

The ClickHouse engine is based on the Snowflake, MySQL, and Exasol implementations. Like Snowflake it lacks unique constraints. Unlike Snowflake it doesn't support primary keys in the traditional sense; they simply enforce physical sort ordering of records.

The engine supports only the latest version of ClickHouse, v25.8, because it requires `UPDATE` and `DELETE`, which were not available in previous versions. The registry schema enables the feature with the required configuration on each table:

```
SETTINGS enable_block_number_column = 1, enable_block_offset_column = 1
```

One wrinkle that had to be worked around: the `clickhouse client` uses the ClickHouse "Native Protocol" on port 9000 by default, while the ODBC driver relies on the HTTP protocol on port 8123 by default. They physically cannot use the same port, so the port specified in a `db:clickhouse` URL cannot be used for the client. Sqitch works around this issue by supporting a custom URL query parameter, `NativePort`, to specify the port for the client to use. Perhaps someday there will be a DBI driver using [clickhouse-cpp](https://github.com/ClickHouse/clickhouse-cpp) and therefor the native protocol. But in the meantime we have to work around this variation between the two connectors.

This implementation is largely complete, but a few things remain to bring it fully in line with the other engines:

- [x] A tutorial (#904)
- [x] Support for parsing the ClickHouse [config file](https://clickhouse.com/docs/interfaces/cli#configuration_files) (#906)
- [x] Configuration file parsing (#906)
- [x] Password-less support (#906)

This last item is difficult, because the behaviors of the `clickhouse client` and the ODBC driver vary considerably.

*   The CLI supports `--jwt` but the [ODBC driver](https://github.com/ClickHouse/clickhouse-odbc) does not
*    The CLI supports SSH authentication with `--ssh-key-file` and `--ssh-key-passphrase` but the ODBC driver does not
*   With `--secure` and a private key specified in a configuration file, the client supports client cert authentication; but the ODBC driver only supports a `PrivateKeyFile` parameter. Sqitch could probably support the former once it learns to parse the config file, but not the latter unless a the client gains a `--private-key` option

At any rate look for those improvements in future commits.

### DIFF
--- a/lib/App/Sqitch/Engine/Upgrade/clickhouse-1.0.sql
+++ b/lib/App/Sqitch/Engine/Upgrade/clickhouse-1.0.sql
@@ -7,6 +7,7 @@ CREATE TABLE releases (
     installer_email TEXT                 NOT NULL
                     COMMENT 'Email address of the user who installed the registry release.'
 ) ENGINE = MergeTree
+  ORDER BY version
   SETTINGS enable_block_number_column = 1, enable_block_offset_column = 1
   COMMENT 'Sqitch registry releases.';
 
--- a/lib/App/Sqitch/Engine/Upgrade/clickhouse-1.0.sql
+++ b/lib/App/Sqitch/Engine/Upgrade/clickhouse-1.0.sql
@@ -7,13 +7,11 @@ CREATE TABLE releases (
     installer_email TEXT                 NOT NULL
                     COMMENT 'Email address of the user who installed the registry release.'
 ) ENGINE = MergeTree
+  SETTINGS enable_block_number_column = 1, enable_block_offset_column = 1
   COMMENT 'Sqitch registry releases.';
 
 -- Add the script_hash column to the changes table. Copy change_id for now.
 ALTER TABLE changes ADD COLUMN script_hash TEXT COMMENT 'Deploy script SHA-1 hash.';
-ALTER TABLE changes MODIFY SETTING
-    enable_block_number_column = 1,
-    enable_block_offset_column = 1;
 UPDATE changes SET script_hash = change_id WHERE TRUE;
 
 -- Allow "merge" events.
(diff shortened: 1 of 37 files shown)


lib/App/Sqitch/Engine/Upgrade/clickhouse-1.0.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE releases (
+    version         REAL COMMENT 'Version of the Sqitch registry.' PRIMARY KEY,
+    installed_at    DATETIME64(6, 'UTC') NOT NULL DEFAULT now64(6, 'UTC')
+                    COMMENT 'Date the registry release was installed.',
+    installer_name  TEXT                 NOT NULL
+                    COMMENT 'Name of the user who installed the registry release.',
+    installer_email TEXT                 NOT NULL
+                    COMMENT 'Email address of the user who installed the registry release.'
+) ENGINE = MergeTree
+  SETTINGS enable_block_number_column = 1, enable_block_offset_column = 1
```
serprex, line 10: fine to only define primary key without ORDER BY?

[docs](https://clickhouse.com/docs/engines/table-engines/mergetree-family/mergetree) make it seem like ORDER BY is required, outside of using a setting which will `ORDER BY tuple()` with suboptimal effects

theory: Yeah, my first day of work I was in a training session where they used `PRIMARY KEY` in most of the examples. Seems to be more or less the same thing as `ORDER BY`. Or at least [it is in the absence of `ORDER BY`](https://clickhouse.com/docs/best-practices/choosing-a-primary-key).

serprex: Primary Key is sparse index, `ORDER BY` is physical sorting of data. Very confusing I know

theory: Queries most often order by date, but deletions are done by primary key. What _should_ it be?

serprex: I think you want both

PR #913, opened 2025-11-10
Add `sqitch grep` command.

Ovid:
After a nail-biting five years of anticipation, here's the pull request for [sqitch grep](https://github.com/sqitchers/sqitch/issues/532).

Full tests and documentation included.

```
Name
    sqitch-grep - Search change scripts and display results in plan order

Synopsis
      sqitch grep [options] <search-term> [<search-term> ...]
      sqitch grep ALTER TABLE
      sqitch grep --type deploy "CREATE INDEX"
      sqitch grep --insensitive --list "drop table"
      sqitch grep --regex "foo.*bar"
```

`sqitch help grep` will give full information.

Disclaimer: while the `grep.pm` code is based on the code I wrote several years ago, the tests and the documentation have been written by [Amazon's Kiro](https://kiro.dev/) using Claude Sonnet 4.5.


theory:
> What does this mean? Do you want me to remove the reviews on this page? Happy to squash to a single commit, but I'm unsure of what you're asking above.

I tried to delete the LLM comments, but got an error; I've reported the bug to GitHub. If you can remove them, great! But I assume this PR is stuck with them.

No, what I meant is that all the generated code needs to be exactly like all the other code in the project, and if it's not, should be removed and replaced or refactored, as appropriate. I trust you, not code generators.

Er, and it sounds like you've already done that. 🤦🏻 I'll review sometime this week, I hope.

Don't worry about the CI failures, BTW; something's broken with the Windows builds and the Snowflake builds don't work for PRs from users without the permissions for the Snowflake credentials. Though I would have thought you'd have them TBH. I'll make sure those tests pass at merged time.

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+use File::Find::Rule;
+use File::Basename        qw(fileparse);
+use File::Spec::Functions qw(catdir splitdir);
+use namespace::autoclean;
+
+extends 'App::Sqitch::Command';
+
+# VERSION
+
+has [qw/type t/] => (
```
theory, line 21: Interesting, I've never seen this syntax before. Too bad it doesn't create just one field and let Getopt::Long map both names to it.

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+        # "old" sqitch files which didn't make it into the plan.
+        # Or, um, there's a bug in my code.
+        return ( $order_by{ $name_for{$a} } // $i ) <=> ( $order_by{ $name_for{$b} } // $i );
+    };
+    @files = sort $by_plan @files;
+
+    if ( $self->list || $self->l ) {
+        say for @files;
+    }
+    else {
```
theory, line 129:
```suggestion
    } else {
```

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+    # sort files by the order in which the names show up in the plan
+    my $by_plan = sub {
+        return 0 if $a eq $b;    # shouldn't happen?
+
+        # if a and b are not exact matches, always return in the order of
+        # deploy, verify, revert
+        return -1 if $a =~ /^$deploy_dir/ and $b !~ /^$deploy_dir/;
+        return -1 if $a =~ /^$verify_dir/ and $b =~ /^$revert_dir/;
+        return 1  if $b =~ /^$deploy_dir/ and $a !~ /^$deploy_dir/;
+        return 1  if $b =~ /^$verify_dir/ and $a =~ /^$revert_dir/;
```
theory, line 90: Should probably escape those `$*_dir` names in the regular expressions.

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+
+        # ok, we got to here. Their top-level directory is the same, so let's
+        # figure out the sort order
+        my $dir;
+        foreach ( $deploy_dir, $verify_dir, $revert_dir ) {
+            $dir = $_ if $a =~ /^$_/;
```
theory, line 96:
```suggestion
            $dir = $_ if $a =~ /^\Q$_/;
```

theory: This will false match on, say `deploy/deploy.sql`, no?

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+
+  my $cmd = App::Sqitch::Command::grep->new(%params);
+  $cmd->execute;
+
+=head1 Description
+
+A lightweight version of C<grep>, this command allows you to search for files
+in your C<sqitch> directories, but it returns them in the order they were
+defined in your plan (with deploy, verify, and revert directories being sorted
+in that order).
```
theory, line 194: If the goal is to search change scripts in plan order, why not use the plan to figure out which files to search, rather than File::Find::Rule?

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+sub show_matches {
+    my ( $self, $files, @args ) = @_;
+    my $pattern = join ' ', @args;
+
+    # Escape regex special chars unless --regex is specified
+    $pattern = quotemeta($pattern) unless ( $self->regex || $self->e );
+
+    my $regex = ( $self->insensitive || $self->i ) ? qr/$pattern/i : qr/$pattern/;
+    FILE: foreach my $file (@$files) {
+        if ( open my $fh, '<:utf8_strict', $file ) {
```
theory, line 143: Do we know for sure that everyone uses UTF-8 in their changes files? I'm less rue. Might be better to use `:raw`.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+##############################################################################
+subtest 'Command construction' => sub {
+    my $grep = App::Sqitch::Command->load(
+        {
+            sqitch  => $sqitch,
+            command => 'grep',
+            config  => $config,
+            args    => [],
+        }
+    );
```
theory, line 95: Closer to project standard style:

```suggestion
    my $grep = App::Sqitch::Command->load({
        sqitch  => $sqitch,
        command => 'grep',
        config  => $config,
        args    => [],
    };
```

Similar formatting changes needed below.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+      },
+      'Should have configuration with all options';
+
+    is_deeply $CLASS->configure( $config, { type => 'verify' } ),
+      { type => 'verify' },
+      'Should configure with type=verify';
+
+    is_deeply $CLASS->configure( $config, { type => 'revert' } ),
+      { type => 'revert' },
+      'Should configure with type=revert';
```
theory, line 127: Project-standard formatting is 4-char indentation for continuing lines, e.g.,:

```suggestion
    is_deeply $CLASS->configure( $config, { type => 'revert' } ),
        { type => 'revert' },
        'Should configure with type=revert';
```

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+
+    is_deeply $CLASS->configure( $config, { list => 1 } ),
+      { list => 1 },
+      'Should configure with list only';
+
+    is_deeply $CLASS->configure( $config, { regex => 1 } ),
+      { regex => 1 },
+      'Should configure with regex only';
+
+    is_deeply $CLASS->configure( $config, { type => 'deploy', insensitive => 1 } ),
```
theory, line 141: I wonder if anyone would want to specify multiple types, as in "search deploy and revert but not verify". Probably not worth the hassle unless someone wants it badly enough.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+      { list => 1, regex => 1 },
+      'Should configure with list and regex';
+};
+
+##############################################################################
+subtest 'Attributes' => sub {
+    my $grep = $CLASS->new( sqitch => $sqitch );
+    isa_ok $grep, $CLASS, 'new grep command';
+    is $grep->type,        undef, 'type should be undef by default';
+    is $grep->t,           undef, 't should be undef by default';
```
theory, line 155: I don't love this duplication, especially since App::Sqitch::Command supports an `options` method in subclasses that map the first string in the option to the attribute name. For example, `bundle.pm` has:

```pl
sub options {
    return qw(
        dest-dir|dir=s
        all|a!
        from=s
        to=s
    );
}
```

So the options automatically map to the attributes named `dest_dir`, `all`, `from`, and `to`. No need for `dirs` and `a` attributes.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        type        => 'deploy',
+        insensitive => 1,
+        list        => 1,
+        regex       => 1,
+    );
+    isa_ok $grep, $CLASS, 'new grep command with options';
+    is $grep->type,        'deploy', 'type should be deploy';
+    is $grep->insensitive, 1,        'insensitive should be 1';
+    is $grep->list,        1,        'list should be 1';
+    is $grep->regex,       1,        'regex should be 1';
```
theory, line 175: Aren't these booleans?

```suggestion
    ok $grep->insensitive,        'insensitive should be 1';
    ok $grep->list,               'list should be 1';
    ok $grep->regex,              'regex should be 1';
```

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+##############################################################################
+subtest 'Literal search functionality' => sub {
+    my $grep = $CLASS->new( sqitch => $grep_sqitch );
+
+    # Test that literal search finds exact strings
+    my @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'CREATE TABLE'
+    );
+    ok scalar(@files) > 0, 'Should find "CREATE TABLE"';
```
theory, line 200: How do we know it found the correct files? Also this doesn't look like Perl. Should be:

```pl
ok scalar @files, 'Should find "CREATE TABLE"';
```

Or:

```pl
ok @files == 1, 'Should find "CREATE TABLE"';
```

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    # Test multiple search terms joined with spaces
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'ALTER', 'TABLE'
+    );
+    ok scalar(@files) > 0, 'Should find "ALTER TABLE" (multiple terms)';
+
+    # Test show_matches output format
+    my @test_files = ( file(qw(t grep-fixtures deploy users.sql))->stringify );
+    my $output     = capture_output { $grep->show_matches( \@test_files, 'CREATE TABLE' ) };
```
theory, line 227: I think you'd be better off using MockOutput to test output, rather than all this fussiness. See, for example, `add.t`.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    like $output, qr/users\.sql:\d+:.*CREATE TABLE/,
+      'show_matches should output filename:line_number:content format';
+    like $output, qr/CREATE TABLE users/,
+      'show_matches should include the matching line content';
```
theory, line 231: This appears to be already covered by the previous test, no?

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+
+    # Test that literal search does NOT match regex patterns
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'foo.*bar'
+    );
+    ok scalar(@files) > 0, 'Should find literal "foo.*bar" string';
```
theory, line 238: Is there any reason not to know the total number of files, and indeed, _which_ files and in what order?

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        'fooXYZbar'
+    );
+    is scalar(@files), 0, 'Should NOT match "fooXYZbar" when searching for "foo.*bar"';
+
+    # Test literal search with period
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'price.'
+    );
+    ok 1, 'Literal search with period completed';
```
theory, line 251: If `get_files` dies, use Test::Exception, instead. It it doesn't what is this test for?

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        sqitch      => $grep_sqitch,
+        insensitive => 1,
+    );
+    @files = $insensitive_grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'create table'
+    );
+    ok scalar(@files) > 0, 'Case-insensitive search should find "create table"';
+
+    # Test literal search with multiple special characters
```
theory, line 277: What special characters?

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+        foreach my $this_file ( $a, $b ) {
+
+            # this takes some time, so cache these puppies
+            unless ( exists $name_for{$this_file} ) {
+                my ( $name, $path, undef ) = fileparse( $this_file, $extension );
+                my @path       = splitdir( catdir( $path, $name ) );
+                my $last_index = -1 * ( @path - scalar @remove );
+                my $this_name  = catdir( splice @path, $last_index );    # strip leading dir
+                $this_name =~ s/\.$//;                                   # remove trailing dot
+                $name_for{$this_file} = $this_name;
```
theory, line 115: ISTM that a lot of this logic could be eliminated in favor of using App::Sqitch::Plan::Change methods like `deploy_file`, and `format_name`.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        regex  => 1,
+    );
+
+    # Test basic regex pattern matching with wildcard
+    my @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'foo.*bar'
+    );
+    ok scalar(@files) > 0, 'Should find pattern "foo.*bar"';
+    like $files[0], qr/(posts|widgets)\.sql$/, 'Should match files with pattern';
```
theory, line 376: Should the oder be deterministic from the plan?

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        '\d+'
+    );
+    ok scalar(@files) > 0, 'Should find pattern with digit class "\d+"';
+
+    # Test regex with word boundary
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        '\bCREATE\b'
+    );
+    ok scalar(@files) > 0, 'Should find pattern with word boundary "\bCREATE\b"';
```
theory, line 439: There is no text in the change scripts with `CREATE` that doesn't have those boundaries, so it doesn't test false positives. I don't think we need to test every regex feature, TBH, but those we do test we should test rigorously.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+
+    # Test invalid regex pattern produces error
+    throws_ok {
+        $grep->get_files(
+            dir(qw(t grep-fixtures deploy)),
+            '(?invalid'
+        );
+    }
+    qr/Unmatched \(|Sequence \(\?/i,
+      'Invalid regex pattern should produce error';
```
theory, line 449: All exceptions should be App::Sqitch::X objects, so all the exception-testing tests should test those objects. Here's an example from `add.t`:

```pl
    throws_ok { $add->_add('sqitch_change_test', $out, $tmpl) }
        'App::Sqitch::X', 'Should get an exception on TT syntax error';
    is $@->ident, 'add', 'TT exception ident should be "add"';
    is $@->message, __x(
        'Error executing {template}: {error}',
        template => $tmpl,
        error    => 'file error - parse error - input text line 1: unexpected end of input',
    ), 'TT exception message should include the original error message';
```

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    ok scalar(@files) > 0, 'Regex case-insensitive search should match mixed case';
+
+    # Test show_matches with regex pattern
+    my @test_files = ( file(qw(t grep-fixtures deploy posts.sql))->stringify );
+    my $output     = capture_output { $grep->show_matches( \@test_files, 'foo.*bar' ) };
+    like $output, qr/foo.*bar/, 'show_matches with regex should display matching pattern';
+
+    @test_files = ( file(qw(t grep-fixtures deploy users.sql))->stringify );
+    $output     = capture_output { $grep->show_matches( \@test_files, '^CREATE' ) };
+    like $output, qr/CREATE TABLE/, 'show_matches with regex anchor should match start of line';
```
theory, line 485: How does this validate that `CREATE TABLE` was matched at the start of aline and not somewhere else? `t/grep-fixtures/deploy/users.sql` has `CREATE TABLE` both at the start of a line and not at the start.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+
+    # Test regex with escaped special characters
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'price\.\$'
+    );
+    ok 1, 'Regex search with escaped special chars completed';
```
theory, line 492: This does not appear to be a valid test unless we're testing that `get_files` didn't raise an error, in which case we should use `lives_ok`.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    my $has_deploy = 0;
+    my $has_verify = 0;
+    my $has_revert = 0;
+    for my $file (@files) {
+        $has_deploy = 1 if $file =~ /deploy/;
+        $has_verify = 1 if $file =~ /verify/;
+        $has_revert = 1 if $file =~ /revert/;
+    }
+    ok $has_deploy,                'No type option should include deploy files';
+    ok $has_verify || $has_revert, 'No type option should include verify or revert files';
```
theory, line 602: Aren't there both?

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        dir(qw(t grep-fixtures deploy)),
+        'CREATE TABLE'
+    );
+    ok scalar(@files) > 0, 'Case-sensitive search should find "CREATE TABLE"';
+
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'create table'
+    );
+    ok scalar(@files) > 0, 'Case-sensitive search should find exact case "create table" in comments';
```
theory, line 635: The fact that it matched files doesn't mean it matched the correct files.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    # Test case-insensitive search with -i flag
+    $grep = $CLASS->new(
+        sqitch      => $grep_sqitch,
+        insensitive => 1,
+    );
+    @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'create table'
+    );
+    ok scalar(@files) > 0, 'Case-insensitive search should find "create table"';
```
theory, line 650: Thi sis no different from the case-sensitive result, so does not test what it purports to test unless it looks at the actual files matched, at the very least.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        sqitch => $grep_sqitch,
+        list   => 1,
+    );
+    is $grep->list, 1, 'List flag should be set';
+
+    my @files = $grep->get_files(
+        dir(qw(t grep-fixtures deploy)),
+        'CREATE TABLE'
+    );
+    ok scalar(@files) > 0, 'List mode grep should find files with pattern';
```
theory, line 686: Doesn't it also find files when not in list mode? How is this different?

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    ok exists $order_by{roles},   'Plan should include roles change';
+    ok exists $order_by{users},   'Plan should include users change';
+    ok exists $order_by{widgets}, 'Plan should include widgets change';
+    ok exists $order_by{posts},   'Plan should include posts change';
+
+    # Verify the plan order is correct
+    is $order_by{roles},   0, 'roles should be first in plan (index 0)';
+    is $order_by{users},   1, 'users should be second in plan (index 1)';
+    is $order_by{widgets}, 2, 'widgets should be third in plan (index 2)';
+    is $order_by{posts},   3, 'posts should be fourth in plan (index 3)';
```
theory, line 798: These appear to test the result of line 786, not the actual execution done in the module. I understand that the algorithm is the same, but if someone changes it in the module this test will not start failing.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+    # Manually sort using the same algorithm as execute()
+    my ( $deploy_dir, $verify_dir, $revert_dir ) = map { $target->$_ } qw/deploy_dir verify_dir revert_dir/;
+    my $extension = $target->extension;
+    my %name_for;
+
+    my $by_plan = sub {
+        return 0 if $a eq $b;
+
+        # deploy < verify < revert for different directories
+        return -1 if $a =~ /^$deploy_dir/ and $b !~ /^$deploy_dir/;
```
theory, line 816: This fails on Windows because the names are not quoted. Also, like above, this test, unless I misread it, tests the logic of code duplicated from the module, not its execution in the module.

t/grep.t
```diff
@@ -0,0 +1,1730 @@
...
+        ok $change_positions{users} < $change_positions{widgets},
+          'users files should appear before widgets files (plan order)';
+    }
+
+    if ( exists $change_positions{widgets} && exists $change_positions{posts} ) {
+        ok $change_positions{widgets} < $change_positions{posts},
+          'widgets files should appear before posts files (plan order)';
+    }
+
+    # Test that files not in plan are sorted last
```
theory, line 925: Why grep files that don't correspond to plan items?

lib/App/Sqitch/Command/grep.pm
```diff
@@ -0,0 +1,315 @@
...
+
+    # Escape regex special chars unless --regex is specified
+    $pattern = quotemeta($pattern) unless ( $self->regex || $self->e );
+
+    my $regex = ( $self->insensitive || $self->i ) ? qr/$pattern/i : qr/$pattern/;
+    FILE: foreach my $file (@$files) {
+        if ( open my $fh, '<:utf8_strict', $file ) {
+            while ( my $line = <$fh> ) {
+                if ( $line =~ /$regex/ ) {
+                    printf "%s:%d: %s" => $file, $., $line;
```
theory, line 146: Please `$self->emit` or one of the other related [output methods](https://github.com/sqitchers/sqitch/blob/42b7192f8ffabfbda79db329f1240a9390458edd/lib/App/Sqitch.pm#L757-L859) that App::Sqitch::Command provides (by delegating to App::Sqitch).

theory (changes requested):
Thanks for this. I stopped reviewing the test after around 920 lines because I figure I've left enough comments already to drive someone mad, and a lot of the issues are the same. Sorry about that.

Overall I like the feature, and want to add it, but it needs some more TLC IMO:

* Consider greping only files associated with changes and ignore others. You can potentially include other categories of files other than deploy/verify/revert by passing any other subdirectories of the top dir to `Change::script_file`; this would fork for [test](https://justatheory.com/2014/01/templating-tests-with-sqitch/) scripts, for example.
* If you really want to grep all the files in those directories, not just the files associated with changes, maybe document it a bit more and explain the rationale. I suspect it's fairly rare, TBH, but am aware of some cases of `\include` --- though I typically recommend putting such files in a separate directory.
* Take advantage of more of the methods in App::Sqitch::Plan::Change if you can. I has all kinds of stuff for normalized names, file paths (and segments), and more.
* A lot of the tests are pretty weak, counting only results rather than inspecting the results, and in some places test nothing or test duplicate code that would not fail if the original code was changed. In other ways it tests more than it needs to! (Honestly just 1-2 tests to show that regular expressions are evaluated at such is enough IMO). I think it needs more massaging to properly evaluate the module.
* Please use the [output methods](https://github.com/sqitchers/sqitch/blob/42b7192f8ffabfbda79db329f1240a9390458edd/lib/App/Sqitch.pm#L757-L859) for all output and then use MockOutput to test output.

theory:
> Also, I listed you as the author on grep.pm. Wasn't sure if there's guidance on that, but happy to have your name there.

Fine with me for you to be listed as author TBH.

PR #927, opened 2026-08-05
attempt to add duckdb as an engine

pmonson711:


etc/templates/deploy/duckdb.tmpl
```diff
@@ -0,0 +1,13 @@
+-- Deploy [% project %]:[% change %] to [% engine %]
+[% FOREACH item IN requires -%]
+-- requires: [% item %]
+[% END -%]
+[% FOREACH item IN conflicts -%]
+-- conflicts: [% item %]
+[% END -%]
+
+BEGIN;
```
theory, line 9: Are DuckDB DMLs transactional?

theory:
Did you use any code generation tools to generate this patch? I haven't published a policy yet, but we will not be accepting patches “written” by “A.I.”

etc/templates/deploy/duckdb.tmpl
pmonson711: Yup, they support transactions.
https://duckdb.org/docs/current/sql/statements/transactions

pmonson711:
> Did you use any code generation tools to generate this patch? I haven't published a policy yet, but we will not be accepting patches “written” by “A.I.”

I did start with opencode, to get much of the code in place. It's been over a decade since I last wrote any Perl :). I totally understand the stance on code generation and even appreciate the policy. I'll close this out.

pmonson711:
@theory I'm happy to put in more manual review and rewrite of this if you are interested in a sans-generative AI implementation. But it's always a bit of a murky thing to remove. Depending on the policy there likely isn't a way to get back to a clean-room, if you will. 

Cheers and thanks for sqitch!

etc/templates/deploy/duckdb.tmpl
theory: Yeah, but do they support them for DML? The only RDMS that does in my experience is Postgres.

theory:
Yes, that would be great, thank you. I agree it can be tough to sort it; code that's copied from elsewhere in Sqitch and hand modified is fine. That's mostly what it takes to add a new engine.

Will need help getting the GH workflows built, too.

etc/templates/deploy/duckdb.tmpl
pmonson711: Sorry if it's a silly question, do you mean transactional DDL? I've always understood DML as inserts, updates, and deletes but perhaps I'm unaware of other topics.

I'll see if I can find details in their documentation but a quick interactive session shows it working with both create tables, indexes and manipulating values.

pmonson711:
Sounds good, I'll drop the tooling and start fresh with comparison of this engine and the others to get it as clean as I can and follow in with my manual testing. I should be able to add to the GH actions as well, but I haven't looked at that setup yet.

etc/templates/deploy/duckdb.tmpl
theory: Not silly at all. I _always_ mix them up. I meant DDL. The test is to start a transaction, create an object, then rollback the transaction, after which the object should not exist.

PR #928, opened 2026-08-23
Require Perl 5.16

theory:
Drop support for Perl 5.10 (long since untested), 5.12, and 5.16. Frankly anything less than 5.18 is a PITA, as demonstrated by the fixes in 6fc9790, but [cpanm stats] show that 5.16 still has about 2% usage (still supported by some older RedHats, as noted in #923), so it seems sensible to keep it for now.

Take advantage of dropping those older Perls to eliminate various workarounds to support them:

*   Fixing the shebang line in `inc/Module/Build/Sqitch.pm`
*   Quoting to avoid File::Copy warnings
*   Finding the home directory on Windows
*   Suppressing a warning in `t/command.t`
*   Using `local $@` in `throws_ok`

  [cpanm stats]: https://cpanmetadb.plackperl.org/versions/


Changes
```diff
@@ -163,7 +164,7 @@ Revision history for Perl extension App::Sqitch
      - Fixed the formatting of the log and plan commands to allow empty or `0`
        separators in lists of things (such as `%{0}t` for a list of tags).
        Thanks to @web-vertalo for the pull request (#703).
-     - Updated the MySQL Tutorial to use 5.7 features. Thanks to Vlad
+     - Updated the MySQL Tutorial to use v5.16 features. Thanks to Vlad
```
autarch: Did you mean to include this in the changes for an earlier verison or the upcoming one?

README.md
```diff
@@ -170,7 +170,7 @@ SOFTWARE.
   [Docker]:    https://img.shields.io/docker/v/sqitch/sqitch?label=%F0%9F%90%B3%20Docker&sort=semver
   [🐳]:        https://hub.docker.com/r/sqitch/sqitch "Latest version on Docker Hub"
   [Perl]:      https://github.com/sqitchers/sqitch/actions/workflows/perl.yml/badge.svg
-  [🧅]:        https://github.com/sqitchers/sqitch/actions/workflows/perl.yml "Tested with Perl 5.12–5.42"
+  [🧅]:        https://github.com/sqitchers/sqitch/actions/workflows/perl.yml "Tested with Perl 16–5.42"
```
autarch: "16-5.42" reads a bit weirdly. I think any of these would be better:

* 5.16-5.42
* 16-42
* 5.16-42

autarch (dismissed):
LGTM % some very minor stuff

Changes
theory: No, over-aggressive find/replace. Fixed in65e6145b, thanks.

README.md
theory: 🤦🏻 Fixed in 65e6145b

